PR #483, opened 2026-01-28
fix: Use valid Perplexity model

exafyltd:
sonar


exafyltd:
Stale (Jan 2026, conflicts with main). conversation.ts and context-pack-builder.ts have evolved substantially. Reopen with fresh branch if Perplexity model fix still needed.

PR #610, opened 2026-04-11
feat(navigator): VTID-NAV-02 — DB-backed catalog + Admin Navigator API

exafyltd:
## Summary

Backs the Vitana Navigator catalog with Supabase so Community Admin users can add/edit the screens + trigger phrases Vitana uses for voice-driven redirects, **without a gateway redeploy**. Ships the backend half of VTID-NAV-02; the React admin screen ships in exafyltd/vitana-v1 on the matching branch.

Root cause this fixes: today `services/gateway/src/lib/navigation-catalog.ts` is a hand-curated static TS constant, so every trigger edit or new screen needs a PR + deploy, and admins can't react to "wrong screen" bug reports. This PR lifts the catalog into DB tables, keeps the static constant as a safety fallback, and wires a secure admin API behind the existing dual-JWT `requireAdminAuth` middleware.

## Schema — `supabase/migrations/20260411000000_nav_catalog.sql`

Three new tables, all RLS-enabled, writes gated at the API layer:

- `nav_catalog` — one row per (screen_id, tenant_id). `tenant_id IS NULL` means shared across all tenants; a non-null tenant row with the same screen_id overrides the shared entry for that tenant at scoring time. Partial unique indexes enforce one-shared + one-per-tenant. New fields on top of the static schema: `context_rules JSONB`, `override_triggers JSONB`, `is_active`, `updated_by`, audit columns.
- `nav_catalog_i18n` — localized `title` / `description` / `when_to_visit` per `(catalog_id, lang)`. `when_to_visit` is the primary trigger text the scorer matches against.
- `nav_catalog_audit` — append-only edit history; every admin write stamps a before/after JSONB snapshot so admins can revert.

## Gateway code

- **`services/gateway/src/lib/nav-catalog-db.ts` (new)** — tenant-keyed in-memory cache with 60 s background refresh + invalidate-on-write. First call falls back to the compile-time `NAVIGATION_CATALOG` constant so the orb never goes dark during deploy / rollback / DB outage. Also exports `findOverrideTriggerMatch` (exact-phrase bypass) and `searchCatalogEntries` — a deliberate copy of `searchCatalog()`'s scorer that operates on a passed-in entry list so the 77 existing navigator tests keep passing unchanged.
- **`services/gateway/src/lib/spa-routes-fallback.ts` (new)** — hard-coded SPA route inventory used by the coverage endpoint until the build-time extract from vitana-v1 ships via CI.
- **`services/gateway/src/routes/admin-navigator.ts` (new)** — 11 endpoints mounted at `/api/v1/admin/navigator`, all gated by `requireAdminAuth` (dual Platform + Lovable JWT via `services/gateway/src/middleware/auth-supabase-jwt.ts`). This is important: the community app signs users into the Lovable Supabase project, so the Platform-only `createUserSupabaseClient` pattern used elsewhere would reject legit community-side admin tokens. The middleware handles both.
  - `GET/POST/PATCH/DELETE /catalog` + `GET /catalog/:id` + `POST /catalog/:id/restore/:audit_id`
  - `POST /simulate` — runs the **real** `consultNavigator()` pipeline against an admin-supplied utterance and returns the full scored result (primary + alternative + top-3 picks with raw scores + decision_source). This is the admin's "verify before save" feedback loop.
  - `GET /spa-routes` — canonical vitana-v1 route list
  - `GET /coverage` — SPA ↔ catalog diff: missing entries, broken catalog routes, dead triggers (30d window via `oasis_events_v1`)
  - `GET /telemetry` — 7/30/90-day aggregates from `orb.navigator.*` events
  - `POST /reload` — force cache refresh (dev convenience)
  - All writes emit audit rows + `invalidateNavCatalogCache()` so the next orb session sees the edit immediately (no 60 s wait).
- **`services/gateway/src/services/navigator-consult.ts`** — three surgical changes:
  1. **Override-trigger shortcut before scoring.** If an admin has registered an exact-match phrase in this tenant's catalog, short-circuit with synthetic high confidence and `decision_source='override_trigger'`. Anonymous sessions still fall through to normal flow so the `anonymous_safe` gate applies.
  2. **Tenant-aware scoring.** `scoreCatalogWithMemory()` prefers the DB cache via `searchCatalogEntries(getCatalogForTenant(...))` when loaded; falls back to the static scorer otherwise. The 77 existing tests exercise the static path and keep passing.
  3. **`top_picks` + `decision_source` on the result.** Top-3 scored entries with raw scores are carried through `NavigatorConsultResult` so the admin Telemetry page can surface near-misses and the Simulator can render the full ranking. `NavigatorConsultPick` gains an optional `score` for the same reason.
- **`services/gateway/src/index.ts`** — imports the new router + mounts it at `/api/v1/admin/navigator` under `owner='admin-navigator'`. Calls `warmNavCatalogCache()` alongside the existing AI personality pre-warm so the tenant cache is populated before the first orb session. The periodic 60 s refresh runs via an unref'd interval inside the warmer — never blocks graceful shutdown.
- **`services/gateway/src/scripts/seed-nav-catalog.ts` (new)** — idempotent standalone script that upserts every current `NAVIGATION_CATALOG` entry as a shared (tenant_id = NULL) row + its i18n children. Run once post-migration to backfill; subsequent runs no-op.

## What does NOT change

- `services/gateway/src/lib/navigation-catalog.ts` is untouched. The static constant still exists as the compile-time fallback and is the data the 77 navigator tests exercise.
- No other existing routes are modified. No existing tests are touched.
- No schema changes to any existing tables.

## Rollback

1. Revert this commit → static catalog serves all requests via the unchanged fallback path (the `nav_catalog` table can sit empty with zero user impact).
2. `DROP TABLE nav_catalog_audit, nav_catalog_i18n, nav_catalog;` if you want a clean slate.
3. No data loss — the static catalog is the source of truth until the seed script runs.

## Test plan

- [ ] `services/gateway/npx tsc --noEmit` clean ✅ (verified locally)
- [ ] `services/gateway/npx jest test/navigation-catalog.test.ts test/navigator-consult.test.ts` — 77/77 passing ✅ (verified locally)
- [ ] Boot-time smoke test: `admin-navigator.ts` module graph loads under tsx with all 11 routes registered ✅ (verified locally)
- [ ] Migration applies cleanly on Supabase (manual or via CI)
- [ ] After merge: run `npx ts-node services/gateway/src/scripts/seed-nav-catalog.ts` with `SUPABASE_URL` + `SUPABASE_SERVICE_ROLE` set → expect `created=~41 updated=0 failed=0`
- [ ] `curl https://gateway-q74ibpv6ia-uc.a.run.app/api/v1/admin/navigator/catalog` (no auth) → 401 UNAUTHENTICATED
- [ ] `curl -H "Authorization: Bearer <admin_jwt>" .../admin/navigator/catalog?tenant_id=__shared__` → 200 with ~41 entries
- [ ] `POST /api/v1/admin/navigator/simulate` with `{ "utterance": "how do I track my biology", "lang": "en" }` → `confidence='high'`, primary screen_id `HEALTH.MY_BIOLOGY`
- [ ] `GET /api/v1/admin/navigator/coverage` → shows real SPA ↔ catalog diff
- [ ] Existing navigator flow unaffected: orb opens, "open my wallet" → `/wallet` (regression)

The React admin UI that consumes these endpoints (`/admin/navigator` under `AdminGuard`) ships in the matching vitana-v1 branch.

### DIFF
--- a/services/gateway/src/index.ts
+++ b/services/gateway/src/index.ts
@@ -221,6 +221,10 @@ if (process.env.K_SERVICE === 'vitana-dev-gateway') {
   const adminTenantsRouter = require('./routes/admin-tenants').default;
   // Admin: Content Moderation
   const adminModerationRouter = require('./routes/admin-moderation').default;
+  // VTID-NAV-02: Admin Navigator — DB-backed catalog CRUD, simulate, coverage, telemetry
+  const adminNavigatorRouter = require('./routes/admin-navigator').default;
+  // VTID-NAV-02: Navigator catalog DB cache warmer (runs at boot)
+  const { warmNavCatalogCache } = require('./lib/nav-catalog-db');
   // Voice Feedback — Test user bug reports & UX improvement suggestions
   const voiceFeedbackRouter = require('./routes/voice-feedback').default;
   // VTID-01250: Autopilot Automations Engine — AP-XXXX registry, executor, wallet, sharing
@@ -561,6 +565,9 @@ if (process.env.K_SERVICE === 'vitana-dev-gateway') {
   // Admin: Content Moderation
   mountRouterSync(app, '/api/v1/admin/moderation', adminModerationRouter, { owner: 'admin-moderation' });
 
+  // VTID-NAV-02: Admin Navigator — catalog/simulate/coverage/telemetry
+  mountRouterSync(app, '/api/v1/admin/navigator', adminNavigatorRouter, { owner: 'admin-navigator' });
+
   // VTID-01250: Autopilot Automations Engine — AP-XXXX registry, executor, wallet, sharing
   mountRouterSync(app, '/api/v1/automations', automationsRouter, { owner: 'automations' });
 
@@ -807,6 +814,14 @@ if (process.env.K_SERVICE === 'vitana-dev-gateway') {
         console.log('🧠 AI Personality config cache pre-warmed');
       } catch (error) {
         console.warn('⚠️ AI Personality cache warm failed (non-fatal, using defaults):', error);
+      }
+
+      // VTID-NAV-02: Pre-warm Navigator catalog DB cache + start periodic refresh
+      try {
+        warmNavCatalogCache();
+        console.log('🧭 Navigator catalog DB cache warming (VTID-NAV-02)');
+      } catch (error) {
+        console.warn('⚠️ Navigator catalog cache warm failed (non-fatal, using static fallback):', error);
       }
     });
   }
--- a/services/gateway/src/lib/nav-catalog-db.ts
+++ b/services/gateway/src/lib/nav-catalog-db.ts
@@ -1,0 +1,505 @@
+/**
+ * VTID-NAV-02: Navigator Catalog — DB cache layer
+ *
+ * Backs navigation-catalog.ts with a live Supabase-sourced catalog so
+ * Community Admin users can add/edit screens and trigger phrases without a
+ * gateway redeploy. The original static NAVIGATION_CATALOG constant remains
+ * as the compile-time fallback: if the DB is down or hasn't been loaded
+ * yet, the orb still routes correctly off the in-code data.
+ *
+ * Design notes
+ * ------------
+ * 1. searchCatalog() / lookupScreen() / entriesByCategory() stay SYNCHRONOUS.
+ *    They read from a module-level cache (Map<tenantKey, NavCatalogEntry[]>).
+ *    Background refreshCatalogCache() keeps the cache fresh. Consumers never
+ *    need to await the catalog.
+ *
+ * 2. The cache is keyed by tenant. Rows with tenant_id IS NULL form the
+ *    SHARED catalog (default for everybody). A row with a non-null tenant_id
+ *    and the same screen_id OVERRIDES the shared row for that tenant only.
+ *    getCatalogForTenant(tid) applies the override.
+ *
+ * 3. First cache fill is triggered from the gateway boot sequence via
+ *    warmNavCatalogCache(). It runs async; failures are logged and swallowed
+ *    because NAVIGATION_CATALOG is already serving requests.
+ *
+ * 4. After an admin write, the admin router calls invalidateNavCatalogCache()
+ *    so subsequent sessions pick up the edit immediately.
+ */
+
+import type { NavCatalogEntry, LangCode, NavCategory } from './navigation-catalog';
+import { NAVIGATION_CATALOG, getContent } from './navigation-catalog';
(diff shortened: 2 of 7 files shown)


exafyltd:
## CI status — pre-existing failures, not introduced by this PR

Two advisory check failures, both reproducible on plain `origin/main` with my branch's changes completely reverted:

1. **Gateway Validation (Minimal CI) — step 12 `Validate Dev frontend navigation spec`** — Fails because `services/gateway/src/frontend/command-hub/navigation-config.js` has drifted from `services/gateway/specs/dev_screen_inventory_v1.json`. Previous commits added tabs to navigation-config.js without updating the spec:
   - `5c0997da fix: integrations & tools — ... add Plugins & Extensions` — added `plugins` to `integrations-tools`
   - `a8f03d44 feat: autonomous self-healing system — ...` — added `self-healing` to `infrastructure`
   - `operator` module also has drifted tabs on main (`dashboard, task-queue, event-stream, deployments, runbook` vs. spec `task-queue, task-details, execution-logs, pipelines, runbook`)
   
   I deliberately did NOT include a spec alignment commit in this PR — the drift spans four unrelated modules and should be resolved in a dedicated cleanup PR so the reviewer can see intent.

2. **Gateway Service Tests — step 7 `Run tests`** — 3 pre-existing failing tests on main, unrelated to the Navigator:
   - `test/task-extractor.test.ts` — `/api/v1/autopilot/health` returning `body.ok = false`
   - `test/autopilot-pipeline.test.ts` — same health endpoint expectation
   - `test/automation-role-targeting.test.ts` — `targetRoles` property missing from registry entries
   
   All three fail on plain `origin/main` (reproduced locally by checking out `origin/main -- .` and running `npx jest test/task-extractor.test.ts test/autopilot-pipeline.test.ts test/automation-role-targeting.test.ts` → `3 failed, 127 passed`). Zero relation to the Navigator feature.

### This PR's own test signals (all green)

- `cd services/gateway && npx tsc --noEmit` → clean
- `npx jest test/navigation-catalog.test.ts test/navigator-consult.test.ts` → **77/77 passing**
- Boot-time module smoke test: `admin-navigator.ts` imports cleanly and registers all 11 routes when loaded under `tsx`
- Full diff across 1342 tests matches main exactly (same 3 pre-existing failures, zero regressions)

The `validate` top-level governance check **passed** (24288892018). The two failing checks above are advisory and the PR's `mergeable_state` is `unstable`, which GitHub documents as "Mergeable but failing some status checks — the user can still merge".

Proceeding with merge. Pre-existing CI drift will be addressed in a separate cleanup PR.

---

PR #943, opened 2026-04-26
DEV-AUTOPILOT: tests for action-executors (5 executors, 0→covered)

exafyltd:
## What

Adds `services/gateway/test/services/action-executors.test.ts` — the first test file for `action-executors.ts`, which was flagged by the missing-tests scanner as having zero coverage.

## Why

`action-executors.ts` registers five executors (`shopping_add_to_list`, `share_milestone`, `social_post_story`, `wearable_log_workout`, `calendar_add_event`) that run after user consent approval. Without tests, regressions in argument coercion, DB error propagation, and return shape correctness are silent until they reach production consent flows.

## Approach

- `jest.mock` captures every `registerActionExecutor(type, fn)` call into an `executors` map so each executor can be invoked directly without going through the full consent gate pipeline.
- `getSupabase` is mocked to return a configurable stub covering both `rpc(...)` and the `from(...).insert(...).select(...).single()` fluent chain.
- `emitClickOutbound` is stubbed to prevent side effects.
- `beforeAll` calls `registerAllActionExecutors()` once; `beforeEach` resets mocks and restores the happy-path stub.

## Files touched

- `services/gateway/test/services/action-executors.test.ts` — **created**

### DIFF
--- a/services/gateway/test/services/action-executors.test.ts
+++ b/services/gateway/test/services/action-executors.test.ts
@@ -1,0 +1,295 @@
+import { registerAllActionExecutors } from '../../src/services/action-executors';
+import type { ActionExecutor, ActionType } from '../../src/services/consent-gate';
+
+// ──────────────────────────────────────────────────────────────
+// Mock: consent-gate — capture registered executors
+// ──────────────────────────────────────────────────────────────
+const executors: Record<string, ActionExecutor> = {};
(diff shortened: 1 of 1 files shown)


exafyltd:
Autopilot v1 e2e — autonomous PR. Demonstrates the autonomy loop end-to-end. CI failed because the auto-generated test had wrong mocks (content-quality issue, not system issue). Closing as part of v1 verification cleanup.

PR #1086, opened 2026-04-29
fix(rls): Enable RLS on oasis_events to deny anonymous access

exafyltd:
This pull request addresses a security finding (`rls_gap`) by enabling Row Level Security (RLS) on the `public.oasis_events` table.

The `oasis_events` table is used by telemetry endpoints and could previously be accessed by any authenticated user, including those with an `anon` role, which poses a security risk.

### Changes

1.  **`supabase/migrations/20251209000000_add_task_stage.sql`**:
    *   RLS is enabled on the `oasis_events` table.
    *   A new policy `deny_unauthenticated` is added to block all `SELECT`, `INSERT`, `UPDATE`, and `DELETE` operations for the `anon` role.

2.  **`services/gateway/test/integration/rls-oasis-events.test.ts`** (new file):
    *   An integration test is added to verify the RLS policy.
    *   It confirms that a client with the `anon` key cannot read from or write to the `oasis_events` table.
    *   It also confirms that a client with the `service_role` key (used by backend services) can still perform all CRUD operations, ensuring no regression in functionality for the telemetry endpoints.

This change hardens the database security by enforcing the principle of least privilege, ensuring that unauthenticated and unauthorized requests are blocked at the database layer.

### DIFF
--- a/services/gateway/test/integration/rls-oasis-events.test.ts
+++ b/services/gateway/test/integration/rls-oasis-events.test.ts
@@ -1,0 +1,140 @@
+import { createClient, SupabaseClient } from '@supabase/supabase-js';
+import { v4 as uuidv4 } from 'uuid';
+import 'dotenv/config';
+
+// Set a longer timeout for integration tests that hit an external service
+jest.setTimeout(30000);
+
+describe('Integration: RLS on oasis_events table', () => {
+  let anonClient: SupabaseClient;
+  let serviceClient: SupabaseClient;
+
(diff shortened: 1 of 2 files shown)


exafyltd:
Closing — modifies an already-applied migration file (`supabase/migrations/20251209000000_add_task_stage.sql` from Dec 2025). Per project standard, applied migrations are append-only. Re-open as a NEW dated migration file (e.g., `supabase/migrations/20260430000000_rls_oasis_events.sql`) that only enables RLS + adds the deny_unauthenticated policy.

PR #1093, opened 2026-04-29
fix(gateway): Correct schema drift in admin-navigator for oasis_events_v1

exafyltd:
This pull request addresses a `schema_drift` violation in `services/gateway/src/routes/admin-navigator.ts` as flagged by the schema drift scanner. The `oasis_events_v1` table's `type` column was renamed to `event_type` in a recent migration, causing queries in this file to become outdated.

This change updates the `GET /coverage` and `GET /telemetry` endpoints to use the correct `event_type` column name in their `SELECT`, `LIKE`, and result processing logic. This restores the functionality of these admin endpoints without any change in their external behavior.

### Files Modified
- `services/gateway/src/routes/admin-navigator.ts`

This is part of a larger effort to fix 60 files affected by the same schema drift. This PR only addresses `admin-navigator.ts`.

### DIFF
--- a/services/gateway/src/routes/admin-navigator.ts
+++ b/services/gateway/src/routes/admin-navigator.ts
@@ -578,8 +578,8 @@ router.get('/coverage', async (req: AuthenticatedRequest, res: Response) => {
     const since = new Date(Date.now() - 30 * 86400000).toISOString();
     const { data: events } = await supabase
       .from('oasis_events_v1')
-      .select('payload, type, created_at')
-      .like('type', 'orb.navigator.%')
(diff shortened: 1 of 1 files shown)


exafyltd:
Closing — based on stale codebase view. PR claims to fix `oasis_events_v1` queries in `admin-navigator.ts` that reference a `type`/`event_type` column. Verified against current `admin-navigator.ts`: no such queries exist; the file's `type`/`topic` references are all for `navigator_screens.related_kb_topics`. Re-open only after verifying the target queries actually exist on current main.

PR #1142, opened 2026-05-01
fix(gateway): tsc compile error breaks every deploy (VTID-02639, DEV-COMHU-2628)

exafyltd:
PR #1137 introduced a TypeScript error in dev-autopilot-execute.ts at line 1247 that has blocked every gateway deploy since 08:30 today.

`lookupR.ok && lookupR.data?.[0]` narrows to `false | { finding_id, pr_url, pr_number }`. Optional chaining only short-circuits on null/undefined, not on false, so `exec?.finding_id` is a compile error when exec is false.

Fix is a one-character coercion: `(lookupR.ok && lookupR.data?.[0]) || undefined`. TS now sees `undefined | { … }` and the optional chain works. Runtime behaviour unchanged.

This unblocks the gateway deploy pipeline for everyone, including PR #1140 which has been waiting to ship the Manuals tab inline render fix.

🤖 Generated with Claude Code

### DIFF
--- a/services/gateway/src/services/dev-autopilot-execute.ts
+++ b/services/gateway/src/services/dev-autopilot-execute.ts
@@ -1243,7 +1243,13 @@ async function patchExecution(
           s,
           `/rest/v1/dev_autopilot_executions?id=eq.${id}&select=finding_id,pr_url,pr_number&limit=1`,
         );
-        const exec = lookupR.ok && lookupR.data?.[0];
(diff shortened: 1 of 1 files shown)


exafyltd:
tsc bug fixed long ago — the patched pattern no longer exists in main (function refactored) and prod deploys are green (gateway-03958-jpm). Superseded.

PR #1228, opened 2026-05-03
DEV-AUTOPILOT: Add test to skip FIXME self-scan in dev-autopilot-scan.mjs

exafyltd:
This PR introduces a unit test (`services/gateway/test/dev-autopilot-scan-self.test.ts`) that verifies the scanner's own file (`scripts/ci/dev-autopilot-scan.mjs`) is skipped during TODO/FIXME scanning. This prevents self-referential false positives caused by the literal string "FIXME" inside the scanner's severity ternary.

**Note to operator:**
The plan also requested modifying `scripts/ci/dev-autopilot-scan.mjs` to add the exclusion guard (`if (relFromRepoLocal(file) === 'scripts/ci/dev-autopilot-scan.mjs') continue;`). However, that file was excluded from the locked file list for this generation, so it could not be modified directly. Please follow up by adding the guard logic inside the `scanTodos` loop in `scripts/ci/dev-autopilot-scan.mjs`. The test has been written with a duplicated logic block inline, as suggested by the plan, to test the behavior of the fix independently.

### DIFF
--- a/services/gateway/test/dev-autopilot-scan-self.test.ts
+++ b/services/gateway/test/dev-autopilot-scan-self.test.ts
@@ -1,0 +1,70 @@
+import fs from 'fs';
+import path from 'path';
+import os from 'os';
+
+// Helper mimicking the behavior from scripts/ci/dev-autopilot-scan.mjs
+function relFromRepoLocal(file: string, repoRoot: string) {
+  return path.relative(repoRoot, file).replace(/\\/g, '/');
+}
+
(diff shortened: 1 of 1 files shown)


exafyltd:
Bulk-closed in the 2026-05-07 sweep — Dev Autopilot regressed and opened 530+ duplicate PRs (root cause: PR #1957). Kill switch is currently armed; nothing will reopen this. If the underlying finding is still valid the autopilot will re-detect it after the guard ships and the kill switch is disarmed.

PR #1238, opened 2026-05-03
DEV-AUTOPILOT: Support request objects for locale in persona greetings

exafyltd:
**Context & Issue:**
During persona handoffs, the voice agent erroneously switches to English even when the user's language is configured otherwise (e.g., German). The root cause is that the caller is passing a context/session object instead of a simple locale string to the greeting lookup. Consequently, `[object Object]` is used as the language key, triggering the default English greeting fallback. Because the system prompt anchors the interaction via the greeting, the LLM continues the entire conversation in English.

**Resolution Plan:**
- Introduce a private helper `extractLocale` in `persona-registry.ts` to normalize language inputs by intelligently inspecting both primitive strings and potential session/context objects.
- Modify `getPersonaGreeting` and `getPersonaGreetingForTenant` to leverage `extractLocale` (and update the type of `lang` to `any` to prevent type mismatches).
- Standardize the extracted locale (e.g., converting `de-DE` into `de`) so that `greeting_templates` dict lookups reliably match the correct template.

**Files modified:**
- `services/gateway/src/services/persona-registry.ts`

### DIFF
--- a/services/gateway/src/services/persona-registry.ts
+++ b/services/gateway/src/services/persona-registry.ts
@@ -98,16 +98,30 @@ export async function getPersonaVoice(key: string): Promise<string> {
   return reg.get(key)?.voice_id ?? '';
 }
 
+function extractLocale(langOrCtx: any): string {
+  let locale = 'en';
+  if (typeof langOrCtx === 'string') {
+    locale = langOrCtx;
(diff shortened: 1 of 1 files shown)


exafyltd:
Superseded by VTID-02694 retry on FB-040

PR #1431, opened 2026-05-04
DEV-AUTOPILOT: Add gateway endpoints for system controls

exafyltd:
This PR adds the necessary Gateway service and route to expose `system_controls` via the API. This is required so the frontend can query flags such as `vitana_did_you_know_enabled` after the associated migration flips the flag to TRUE.

### Changes
* Created `SystemControl` type definition.
* Added `getSystemControl` service function to fetch from the Supabase `system_controls` table.
* Added an Express route at `GET /:key` to serve the system control status.
* Unit tests for both the service and the route.

**Operator Note regarding Command-Hub Frontend:**
The plan implies modifying the command-hub frontend to consume this new endpoint (`GET /api/system-controls/vitana_did_you_know_enabled`). Since the exact frontend file path wasn't provided in the strict locked list, its modification has been omitted from this PR. Please follow up by adding the API call to the command-hub tour component.

### DIFF
--- a/services/gateway/src/routes/system-controls.ts
+++ b/services/gateway/src/routes/system-controls.ts
@@ -1,0 +1,22 @@
+import { Router, Request, Response } from 'express';
+import { getSystemControl } from '../services/system-controls';
+
+const router = Router();
+
+router.get('/:key', async (req: Request, res: Response) => {
+  try {
+    const key = req.params.key;
+    const control = await getSystemControl(key);
+
+    if (!control) {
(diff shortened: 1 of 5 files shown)


exafyltd:
Bulk-closed in the 2026-05-07 sweep — Dev Autopilot regressed and opened 530+ duplicate PRs (root cause: PR #1957). Kill switch is currently armed; nothing will reopen this. If the underlying finding is still valid the autopilot will re-detect it after the guard ships and the kill switch is disarmed.

PR #1914, opened 2026-05-06
docs(maxina): match-card Vertex AI Imagen fallback pipeline spec + shared types

exafyltd:
## Summary

Companion to `exafyltd/vitana-v1#claude/vertex-ai-fallback-images-eZNgU`, which adds the actual edge functions, supabase migration and frontend pieces.

- `specs/maxina-match-image-fallback.md` — full architecture: priority ladder, DB schema, resolver/generator edge functions, cold-start sequence, and the four explicit guarantees (generate once, no shared cartoon, no stranger-as-user, image source always tracked).
- `packages/shared-types/maxina-match-image.ts` — canonical TypeScript contract for the resolver/generator request and response shapes; both repos import the same definitions so any future cache-warming or proxy service stays in lockstep.

## Test plan
- [ ] Spec reads end-to-end as a single source of truth for the pipeline.
- [ ] Shared types compile and match the resolver/generator implementations on `vitana-v1`.
- [ ] Backend services that later need to fan out generation (e.g. nightly cache warmer) can import from `packages/shared-types/maxina-match-image.ts` without re-declaring the enums.

https://claude.ai/code/session_01R57QSgbZiVAgDphiKBYNR9

---
_Generated by [Claude Code](https://claude.ai/code/session_01R57QSgbZiVAgDphiKBYNR9)_


exafyltd:
Closing: docs-only spec/shared-types for a match-card Vertex AI Imagen fallback pipeline, companion to `vitana-v1#416` — neither side of this pair ever got real implementation code in the 3+ months since, so this looks like an abandoned direction rather than in-progress work. Separately, the platform's image-generation fallback strategy has since moved to Amazon Titan/Bedrock (CLAUDE.md §2d, VTID-03497) for the two Imagen consumers that did ship. If match-card image fallback is still wanted, it should be scoped against the current Titan adapter rather than this Vertex-era spec. Leaving `vitana-v1#416` for a separate look since it's a different repo.

---

PR #2015, opened 2026-05-09
fix(orb): reset silent-reregister budget on successful audio send after re-register (VTID-02034c, Codex P2 follow-up to #2013)

exafyltd:
## Summary

Follow-up to **#2013 (VTID-02034b)** addressing Codex's P2 review finding.

[Codex review on #2013](https://github.com/exafyltd/vitana-platform/pull/2013#discussion_r3213022046):

> Because this counter is only reset by the 30s wall-clock window and not when `_resetAndReconnect()` successfully starts a new session, two legitimate cross-instance 404s consume the whole budget even if both silent re-registers worked. In the same multi-instance/no-affinity environment this patch targets, a third audio POST routed to an instance without the newly minted session within 30s will immediately play the user-visible network disconnect again, so the loop can still recur during normal conversation; reset the budget on successful reconnect or after the first successful audio send for the new session.

## Fix

In `_sendAudio`'s `r.ok` branch (the existing success path that already zeroes `_audioSendFailCount`), also zero `_silentReregisterCount` + `_silentReregisterWindowStart` when the count is non-zero. A successful audio POST after a silent re-register is proof that the new session is healthy; the budget should reflect that we paid our way back to a working state rather than holding the count against future legitimate cross-instance 404s during the same conversation.

```diff
       if (r.ok) {
         _s._audioSendFailCount = 0;
+        // VTID-02034c: ... (Codex P2)
+        if (_s._silentReregisterCount > 0) {
+          _s._silentReregisterCount = 0;
+          _s._silentReregisterWindowStart = 0;
+        }
         return;
       }
```

Net change: **+13 lines, -1 line** (1 line is the `_WIDGET_VERSION` bump).

## Why this is correct

1. **It's not the first 404 that's risky** — it's a routine cross-instance 404 that the carve-out is meant to handle. If the carve-out worked (next POST → 200), the recovery worked.
2. **The 30s window was meant for "every instance returning 404"** (real degenerate state) — not for normal cross-instance shuffle that happens to pile up two re-registers within 30s.
3. **Real degenerate state still surfaces correctly**: if the new session also gets 404s back-to-back without ever succeeding, the count stays >0, the budget exhausts as designed, and `_announceDisconnect('network')` fires.

## ⚠️ Apply patch locally before review (same as #2013)

```sh
git fetch origin claude/fix-orb-budget-reset-1QHXq
git checkout claude/fix-orb-budget-reset-1QHXq
git apply VTID-02034c-budget-reset.patch
git rm VTID-02034c-budget-reset.patch
git add services/gateway/src/frontend/command-hub/orb-widget.js
git commit -m "fix(orb): reset silent-reregister budget on successful audio send after re-register (VTID-02034c, Codex P2)"
git push
```

`_WIDGET_VERSION` is bumped to `2026-05-09-vtid-02034c-budget-reset`. The vitana-v1 cache-buster in `index.html` (PR vitana-v1#444) currently points at `?v=20260508-vtid-02034b-stale-instance-404`. Once this lands, bump that query string to `?v=20260509-vtid-02034c-budget-reset` so deployed clients pull the corrected widget.

## Test plan

- [ ] Apply patch + push (instructions above)
- [ ] Wait for gateway deploy
- [ ] Bump vitana-v1 cache-buster to `?v=20260509-vtid-02034c-budget-reset`, deploy v1
- [ ] Hard-refresh, open the orb. Verify `[VTOrb] Widget version: 2026-05-09-vtid-02034c-budget-reset` in console
- [ ] Long-running conversation: trigger several cross-instance 404s spaced over 30+s. Verify `silent re-register #1/2`, then a successful chunk, then `silent re-register #1/2` again (counter visibly resets after the success). No "internet issues" TTS.
- [ ] Pathological case: trigger 3+ consecutive 404s with no success between them. Verify the budget still exhausts → existing `_announceDisconnect('network')` flow still runs as designed.

## Why not also fix the root cause here

The proper fix (shared `liveSessions` store — Redis/Supabase) is still the bigger work item VTID-02036's revert note flagged. This carve-out + Codex's P2 fix together get the user-visible loop to a place where it doesn't surface during normal conversation. Doing the shared-store migration is out of scope for the hot-fix line.

https://claude.ai/code/session_01KDHXsZLNPM2Gcst5j9DrEz

---
_Generated by [Claude Code](https://claude.ai/code/session_01KDHXsZLNPM2Gcst5j9DrEz)_


exafyltd:
Not mergeable as clean: the only file is VTID-02034c-budget-reset.patch (a patch file committed to the repo root) — the actual budget-reset fix was never applied to orb code. Close; re-open as a real code change if still wanted.

PR #2172, opened 2026-05-17
ops: diagnostic workflow — autopilot Job runtime visibility

exafyltd:
One-shot manual workflow that dumps the gateway service env vars, autopilot-executor Job state + IAM, and recent Job executions. Needed to root-cause why autopilot executions get stuck despite VTID-02703 Job runtime being enabled. Workflow-only change, no runtime impact.


exafyltd:
Kitchen-sink diagnostic branch (80 files, conflicting). Reopen scoped if the diagnostic workflow is still needed.

PR #2309, opened 2026-05-22
fix(chat): RLS — let DM recipients & group members view image attachments

exafyltd:
## Summary

Companion to [vitana-v1#514](https://github.com/exafyltd/vitana-v1/pull/514).

Adds the same migration to this repo so the existing **RUN-MIGRATION** workflow (Actions → "Run SQL Migration") can apply it via `SUPABASE_DB_URL`. The migration must live here because `chat_group_members` is defined in this repo and the workflow operates on this repo's checkout.

The migration extends the `chat-attachments` bucket's SELECT policy with two new arms so DM recipients and chat_groups members can re-sign URLs to view images uploaded by other users (current behavior: only the uploader can, recipients get a broken-image icon).

See vitana-v1#514 for the full root-cause writeup.

## How to apply

1. Merge this PR (or leave as draft — the workflow can run off this branch directly).
2. Actions → **Run SQL Migration** → Run workflow.
3. `migration_file`: `supabase/migrations/20260522000000_chat_attachments_rls_dm_and_groups.sql`
4. Branch: `claude/chat-image-upload-fix-cNdyh` (or `main` after merge).

## Test plan

- [ ] Workflow runs to completion (psql apply + PostgREST schema reload)
- [ ] User A sends an image DM to User B → B sees the image
- [ ] Member uploads an image in 🎆 FIRST 100 → other group members see it
- [ ] Non-group-member can't `createSignedUrl` for that group's attachment paths
- [ ] Existing 1:1 DMs older than 1h regain images for both participants

---
_Generated by [Claude Code](https://claude.ai/code/session_016n84ibVegqhgQmBW1XCCNc)_

### DIFF
--- a/services/gateway/src/routes/chat.ts
+++ b/services/gateway/src/routes/chat.ts
@@ -36,16 +36,37 @@ router.post('/send', requireAuth, requireTenant, async (req: Request, res: Respo
   const { identity } = req as AuthenticatedRequest;
   if (!identity) return res.status(401).json({ ok: false, error: 'unauthorized' });
 
-  const { receiver_id, content } = req.body;
+  const { receiver_id, content, message_type, content_data } = req.body as {
+    receiver_id?: unknown;
+    content?: unknown;
+    message_type?: unknown;
+    content_data?: unknown;
+  };
 
   if (!receiver_id || typeof receiver_id !== 'string') {
     return res.status(400).json({ ok: false, error: 'receiver_id is required' });
   }
-  if (!content || typeof content !== 'string' || content.trim().length === 0) {
-    return res.status(400).json({ ok: false, error: 'content is required' });
-  }
   if (receiver_id === identity.user_id) {
     return res.status(400).json({ ok: false, error: 'cannot message yourself' });
+  }
+
+  const msgType = typeof message_type === 'string' && message_type.length > 0 ? message_type : 'text';
+  const allowedTypes = new Set(['text', 'attachment', 'voice', 'voice_transcript']);
+  if (!allowedTypes.has(msgType)) {
+    return res.status(400).json({ ok: false, error: 'invalid_message_type' });
+  }
+
+  const rawContent = typeof content === 'string' ? content : '';
+  const trimmedContent = rawContent.trim();
+  const metadata = content_data && typeof content_data === 'object' ? (content_data as Record<string, unknown>) : {};
+  const attachments = Array.isArray((metadata as any).attachments) ? (metadata as any).attachments : [];
+
+  if (msgType === 'text') {
+    if (trimmedContent.length === 0) {
+      return res.status(400).json({ ok: false, error: 'content is required' });
+    }
+  } else if (trimmedContent.length === 0 && attachments.length === 0) {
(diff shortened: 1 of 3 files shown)


exafyltd:
Found a second, deeper bug while testing — added a follow-up migration.

The user hit `infinite recursion detected in policy for relation "chat_group_members"` when sending an image **DM** (not even a group message). Root cause is in the `chat_groups` migration (`20260525000000_VTID_03089_chat_groups.sql`): the `chat_group_members_read` SELECT policy does `EXISTS … FROM chat_group_members …` against the same table — Postgres re-applies the same policy on the inner query and tries to recurse forever.

It surfaces on DM sends because the same migration added a `users_read_group_messages` SELECT policy on `chat_messages` that also calls `EXISTS … FROM chat_group_members …`. The v1 client's `useGlobalMessages.sendMessageLegacy` uses `.insert().select()` on `chat_messages`, which evaluates all SELECT policies on the inserted row — including the group one, even when `group_id IS NULL`. The inner subquery trips the chat_group_members recursion before the row is returned.

New migration `20260522010000_chat_group_members_recursion_fix.sql`:
- Adds `is_chat_group_member(group_id, user_id)` as `SECURITY DEFINER STABLE` so the membership check bypasses RLS.
- Rewrites `chat_group_members_read`, `chat_groups_member_read`, `users_read_group_messages` and `users_send_group_messages` to call the helper.
- Rewires the `chat_groups` arm of the `chat-attachments` storage policy (from this PR's first migration) to use the helper too, and guards the `path[2]::uuid` cast against non-UUID second folders so a stray legacy object can't blow up the policy.

Apply order via RUN-MIGRATION: `20260522000000_chat_attachments_rls_dm_and_groups.sql` first, then `20260522010000_chat_group_members_recursion_fix.sql` (the latter is also self-contained and recreates the storage policy in full, so applying just the second one also works).

---

exafyltd:
User confirmed text DMs work but image DMs still fail with the same recursion error → root cause narrowed to the v1 client's split send path. Plain text goes through the gateway (`/api/v1/chat/send`, service_role insert, RLS bypassed), but attachment / voice DMs did a client-side `supabase.from('chat_messages').insert().select()` as the authenticated user — which hits the recursive `chat_group_members` policy via `users_read_group_messages` and fails before the row commits.

New commits provide a code-side workaround that doesn't depend on the SQL migration running first:

- **gateway (this PR, `7dc8bb1`)** — `POST /api/v1/chat/send` now accepts `message_type` + `content_data`, mirroring the `chat_groups /:id/send` shape. Persists `message_type` + `metadata` and uses "📎 \<filename>" / "🎤 Voice message" in the push notification body for attachment-only / voice-only DMs. Skips the @vitana auto-reply when the message has no text prompt.
- **vitana-v1#514 (`457f7af`)** — `sendChatMessage` accepts `{ messageType, contentData }`; `useGlobalMessages.sendMessageLegacy` routes all DM types through the gateway (the Supabase direct-insert path stays as a gateway-down fallback but now carries the rich-content fields too).

This makes the send work today even without the recursion-fix migration applied; the SQL migration becomes defense-in-depth for any future code path that ever does a direct insert as the user.

---

PR #2432, opened 2026-05-31
[FleetView] feat(orb): DEV-COMHU-0502 reactive auth contract + identity telemetry (ORB Recovery 1)

exafyltd:
## Summary
ORB Recovery **Phase 1 — Auth contract** (the biggest UX lever). The widget's `setAuth` was a no-op once `init()` ran without a token (`forceAnonymous`/`_tokenSetByInit` permanent gate). A host that initialised the widget *before* login resolved, then called `setAuth(token)` on login, **stayed anonymous forever** — silently skipping memory, cadence, last-session info, and authenticated tools. That one bug explains the "I have no access" responses **and** missing-memory **and** the "first-time greeting on every reopen" cluster.

## What this PR ships
**Widget (`orb-widget.js`)**
- `setAuth(token)` is now **reactive**: a non-empty token always lifts the anonymous lock (`forceAnonymous = false`) and takes ownership of auth. Empty/null routes to `clearAuth()` (logout semantics).
- New **`clearAuth()`** — clears the token AND identity-bound continuity (`_transcriptHistory`, `conversationId`, `_preDisconnectStage`, `_reconnectCount`). This **preserves the VTID-AUTH-FIX anti-leak guarantee** ("Hello Jovana/Dragan" bug) on logout/account-switch — the prior `forceAnonymous` lock was a blunt instrument; `clearAuth` is the precise one.
- Cache-bust → `20260531-DEV-COMHU-0502-auth-contract`.

**Gateway**
- New typed OASIS topic `orb.session.identity.resolved`.
- `handleLiveSessionStart` emits it (fire-and-forget) with `has_authorization_header`, `auth_valid`, `is_anonymous`, `surface`, and an explicit **`anonymous_on_authenticated_surface`** flag — the exact metric the Phase D cockpit uses to detect widget-side anonymous drift from the outside.

## Deliberate deviation from the plan snippet (and why)
The plan suggested "remove the `_tokenSetByInit` permanent gate" and "refuse to start anonymous on an authenticated surface (401)". I did **not** do those verbatim, because:
1. Naively removing the gate would **reintroduce** the VTID-AUTH-FIX identity-leak bug (stale localStorage token leaking a previous user). Instead, `setAuth` is reactive *and* `clearAuth` wipes continuity — same UX outcome, anti-leak preserved.
2. A blanket 401-refuse-anonymous on community/admin would break **legitimately anonymous** public-landing sessions the gateway supports by design. The identity event makes the drift **observable** first; a hard refusal (if wanted) should ship behind a flag after the cockpit confirms the drift is gone. Flagged for your call — happy to add the gated refusal if you want it.

## Tests (green locally)
- `npm run typecheck` ✓, `npm run build` ✓, `npm run smoke:orb-widget` ✓
- `npx jest test/frontend/` ✓ (incl. 3 new reactive-auth static checks)

## Vertex parity ✓
Identity event fires on `handleLiveSessionStart` (the Vertex session-start path). Widget auth contract is upstream-agnostic.

## LiveKit parity ✓ (with companion patch)
`setAuth`/`clearAuth` are provider-agnostic and ship via the shared gateway-served widget. The host-side wiring (`useOrbVoiceClient.ts` + `useLiveKitVoice.ts` reactive `setAuth`, shell `clearAuth()` on logout) is documented in `docs/patches/vitana-v1/ORB-1-auth-contract.md`.

## Pending human actions (aggregation file)
- Merge; EXEC-DEPLOY SUCCESS; `/alive` 200.
- Apply `docs/patches/vitana-v1/ORB-1-auth-contract.md`.
- Acceptance: login dragan3 → `is_anonymous=false` + memory/cadence; logout → no silent authenticated drift; dragan1↔dragan3 switch → no leak; both providers.
- Decide whether to add the flag-gated authenticated-surface refusal.

---
🤖 ORB Recovery 1 per the autonomous-execution plan. Sandbox cannot merge/deploy/run prod smokes.

---
_Generated by [Claude Code](https://claude.ai/code/session_01ApimmhkZML398iKgkgVSPC)_

### DIFF
--- a/docs/patches/vitana-v1/ORB-1-auth-contract.md
+++ b/docs/patches/vitana-v1/ORB-1-auth-contract.md
@@ -1,0 +1,71 @@
+# vitana-v1 companion patch — ORB Recovery 1: Auth contract (DEV-COMHU-0502)
+
+**Target repo:** `exafyltd/vitana-v1` (NOT reachable from the autonomous sandbox)
+**Companion to:** vitana-platform PR — reactive `setAuth` + `clearAuth()` in
+`services/gateway/src/frontend/command-hub/orb-widget.js`, and the
+`orb.session.identity.resolved` OASIS event in `live-session-controller.ts`.
+
+## Why
+The widget now treats `setAuth(token)` as **reactive**: a real token lifts the
+anonymous lock at any time, and `setAuth('')`/`clearAuth()` wipes identity-bound
+continuity. For this to fix the "I have no access / missing memory / first-time
+greeting every reopen" cluster, the host shell must actually CALL `setAuth` on every
+token-state change and `clearAuth` on logout/account-switch.
+
+## 1. Cache-bust bump (required)
+```diff
+- orb-widget.js?v=20260529-VTID-03185-audio-queue-closure
++ orb-widget.js?v=20260531-DEV-COMHU-0502-auth-contract
+```
+
+## 2. `src/hooks/useOrbVoiceClient.ts` — call setAuth reactively
+React to Supabase session changes (login, silent refresh, account switch) and push the
+fresh access token into the widget. Example:
+
+```ts
+import { useEffect } from 'react';
+import { supabase } from '../lib/supabaseClient';
+
+export function useOrbVoiceClient() {
+  useEffect(() => {
+    // Push the current token immediately…
+    supabase.auth.getSession().then(({ data }) => {
+      const token = data.session?.access_token;
+      if (window.VitanaOrb) {
+        token ? window.VitanaOrb.setAuth(token) : window.VitanaOrb.clearAuth();
+      }
+    });
+    // …and on every subsequent change.
+    const { data: sub } = supabase.auth.onAuthStateChange((event, session) => {
+      if (!window.VitanaOrb) return;
+      if (event === 'SIGNED_OUT' || !session?.access_token) {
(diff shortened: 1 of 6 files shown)


exafyltd:
Both Codex P1 findings were correct and are fixed in `0527ce8`:

**1. `setAuth` account-switch leak (line ~2794).** Wiping continuity on *every* `setAuth` would break the same-user silent-refresh case this PR exists to support, so the fix is identity-aware: a new `_jwtSub(token)` compares the JWT `sub`. On a **sub change** (account switch) `setAuth` now wipes `_transcriptHistory`/`conversationId`/`_preDisconnectStage`/`_reconnectCount` **and** stops the old-identity live session before the next `_sessionStart`. Same-user refresh (sub unchanged) keeps continuity. `init()` seeds `_lastAuthSub` so the first post-init `setAuth` isn't misread as a switch.

**2. `clearAuth` left the live session running (line ~2806).** `clearAuth` now calls `_sessionStop()` **first** (while the old token is still set, so the stop authenticates as the departing identity), then clears the token + continuity. Post-logout/switch turns can no longer be handled or persisted against the prior `session.identity`.

Shared `_wipeIdentityBoundState()` helper; jest extended (account-switch-only reset, logout teardown ordering, `_jwtSub`, init baseline). typecheck + smoke + widget suites green; cache-bust → `…-auth-contract-r2`.

---

PR #2438, opened 2026-05-31
[FleetView] feat(orb): DEV-COMHU-0505 executable autopilot CTA contract (ORB Recovery 5)

exafyltd:
## Summary
ORB Recovery **Phase 5 — autopilot CTA contract**. The autopilot recommendation next-action generated `CTA type='ask_permission'` with a `recommendation_id` but **no tool wired**. When the user said "yes", the model had no deterministic action and either called the wrong tool or answered **"I have no access"** — even though `activate_recommendation` already exists and works.

## Root cause (and what was already fine)
- The tool **already exists and is shared**: `tool_activate_recommendation` in `orb-tools-shared.ts`, registered in `live-tool-catalog.ts`, and it already returns distinguishable outcomes (`activated` / `already_active` / `recommendation_not_found` / `recommendation_belongs_to_another_user`). So the "truthful fallback (not 'I have no access')" requirement is **already satisfied** at the tool layer — no change needed there.
- The gap was purely the **CTA contract**: the offer didn't tell the model *which* tool to call on "yes".

## What this PR ships
- `ContinuationCta`'s `ask_permission` variant gains an optional **`onYesTool`** (back-compatible — other sources omit it).
- `autopilot-recommendation.ts` attaches `onYesTool: 'activate_recommendation'` + `payload.id` (and keeps `recommendation_id`), so "yes" deterministically invokes the shared activator with the right id.
- CTA-contract unit test added.

## Scope note (honest)
The plan also asks to write the **pending CTA into `orb_session_state`** (so the accepted "yes" resolves cross-transport even if the model loses it). That needs session-id plumbing into the continuation path + live verification — documented as the follow-up. The CTA-contract fix here is the part that directly removes the "I have no access" mismatch and is fully testable.

## Tests (green locally)
- `npm run typecheck` ✓, `npm run build` ✓
- `npx jest .../autopilot-recommendation.test.ts` ✓ (12, incl. new CTA-contract case)

## Vertex parity ✓
CTA flows through the shared continuation stack both providers consume; the shared activator is unchanged.

## LiveKit parity ✓ (with companion patch)
`docs/patches/orb-agent/ORB-5-autopilot-cta.py` — agent declares `activate_recommendation` with the identical signature (delegating to the shared REST activator) + prefers the pending CTA on an affirmative turn.

## Pending human actions (aggregation file)
- Merge; EXEC-DEPLOY SUCCESS; `/alive` 200.
[...]

PR #2435, opened 2026-05-31
[FleetView] feat(orb): DEV-COMHU-0503 close/reopen continuity + cadence writer (ORB Recovery 2+3)

exafyltd:
## Summary
ORB Recovery **Phase 2+3 — close/reopen continuity + cadence** (combined PR). Two root causes:
1. `_hide()` tore down all continuity, so closing the ORB for a minute looked **"first-time"** on reopen.
2. `wake_cadence:last_turn_at` had a **reader but no writer** — so the greeting-decay policy always saw "no prior turn" and the same wake line could fire every reopen. (VTID-03172 tried to fix this at the prompt layer — wrong layer.)

## What this PR ships
**Backend**
- **`recordWakeTurn()`** — the missing `wake_cadence:last_turn_at` writer (call on every meaningful turn).
- **Migration** `20260606000000_DEV_COMHU_0503_orb_session_state.sql` — TTL'd shared cross-transport state table (also backs ORB-4/ORB-5). **Committed as a file; not run from the sandbox.**
- **`orb-session-state.ts`** — typed `read`/`write`/`clear` helper, fail-open reads, TTL by close reason.
- **Continuity endpoints** `POST`/`GET`/`DELETE /api/v1/orb/session/continuity` (authenticated; anonymous returns `ok:false` — no durable identity).

**Widget (`orb-widget.js`)**
- `_persistContinuity(reason, ttl)` via `keepalive` POST; `_hide()` now persists 15-min continuity **before** `_sessionStop` teardown (teardown itself unchanged → low regression risk).
- `_clearContinuity()` + new public **`reset()`** for logout / account-switch (clears durable + in-memory identity-bound state — anti-leak).
- Cache-bust → `20260531-DEV-COMHU-0503-continuity`.

## Scope note (honest)
The deepest part of the plan — rewriting `handleLiveSessionStart` to *hydrate* `session.conversationId`/`transcriptHistory`/`lastTurnAt` from `orb_session_state` and refactor `decideGreetingPolicyAuthoritative` — is the piece that genuinely needs live-session testing I can't do in the sandbox. This PR lands the **durable substrate** (table + helper + endpoints + writer + widget persistence) that hydration builds on; the hydration wiring + greeting-policy refactor is captured in the patch files for a follow-up with real-session verification. Flagging so it's not mistaken for complete.

## Tests (green locally)
- `npm run typecheck` ✓, `npm run build` ✓
[...]

PR #2488, opened 2026-06-01
[FleetView] fix(orb): activate_recommendation reads pending_cta on bare "yes" (ORB-5 runtime follow-up)

exafyltd:
## Summary
Runtime follow-up to ORB Recovery 5 (#2438) — closes the **read side** Codex flagged. #2438 made the autopilot CTA *carry* `onYesTool` and persisted it to `orb_session_state` (`pending_cta`). This wires the **consumer**: when the model calls `activate_recommendation` after a spoken "yes" with no `id` in context, the tool now resolves the id from the persisted pending CTA instead of dead-ending on `"id is required"` (the "I have no access" symptom).

## What this PR ships
- `tool_activate_recommendation` (`orb-tools-shared.ts`): if `args.id` is empty and the user is authenticated, read `orb_session_state` key `pending_cta`; if it holds `{ tool: 'activate_recommendation', payload.id }`, use that id and **consume the row** (`clearOrbSessionState`) so a later turn can't re-activate a stale offer. Best-effort — any failure falls through to the existing `"id is required"` guard.
- Used by the shared dispatcher, so **both Vertex and the LiveKit tool wrapper** get it.

## Tests
- `voice-activate-recommendation-shared.test.ts`: +2 cases — (6b) empty id resolves the rec from `pending_cta` and consumes it; (6c) empty id + no `pending_cta` still returns `"id is required"`. Full suite **9/9**. typecheck + build green.

## Vertex parity ✓ / LiveKit parity ✓
Pure shared-dispatcher logic; no provider-specific path. The LiveKit agent reads `pending_cta` per `docs/patches/orb-agent/ORB-5-autopilot-cta.py`.

## Dependency
Reads `orb_session_state` (helper from #2435). **Requires the `orb_session_state` migration applied** to function; without it, `readOrbSessionState` returns null and behavior degrades to the prior `"id is required"` — no regression.

---
🤖 ORB Recovery 5 runtime follow-up. Sandbox cannot merge/deploy/run prod smokes — but this PR is pure code + unit-tested.

---
_Generated by [Claude Code](https://claude.ai/code/session_01ApimmhkZML398iKgkgVSPC)_

### DIFF
--- a/services/gateway/src/services/orb-tools-shared.ts
+++ b/services/gateway/src/services/orb-tools-shared.ts
@@ -2107,7 +2107,36 @@ export async function tool_activate_recommendation(
   id: OrbToolIdentity,
   sb: SupabaseClient,
 ): Promise<OrbToolResult> {
-  const recId = String(args.id ?? '').trim();
+  let recId = String(args.id ?? '').trim();
+  // DEV-COMHU-0505 (review follow-up): when the model calls this after a spoken
+  // "yes" it often has no id in context — the offer's id lives in the persisted
+  // pending CTA (written by wake-brief-wiring into orb_session_state). Fall back
+  // to it so the affirmative turn resolves deterministically instead of erroring
+  // with "id is required" (the "I have no access" symptom). Authed users only.
+  if (!recId && id.user_id) {
+    try {
+      const { readOrbSessionState, clearOrbSessionState } = await import('./orb/orb-session-state');
+      const pending = await readOrbSessionState<{ tool?: string; payload?: { id?: string } }>(
+        sb,
+        id.user_id,
+        'pending_cta',
+      );
+      const pendingId =
+        pending &&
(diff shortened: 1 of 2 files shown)


exafyltd:
Valid P2 — fixed in `cb97f99`. Clearing `pending_cta` on read meant a transient fetch/update error after the fallback would drop the user's only recovery context, turning a retryable "yes" into a permanent "id is required". The consume now happens **only on the success path**, after the recommendation is verified + activated (fire-and-forget). Test 6b drains microtasks before asserting the deferred delete; 6c (no pending CTA → still "id is required") unchanged.

Separately: the earlier `Gateway Service Tests` red on `0332451` was a flake — I re-ran the **full** gateway suite locally (287 suites / 4946 tests, 0 failures). This push re-triggers CI fresh.

---

PR #2786, opened 2026-06-25
feat(vcaop): Admitad product catalog sync (Products API → /discover)

exafyltd:
## What

Adds **Admitad** as a marketplace **catalog** source. Admitad was previously wired as a rewards/postback network only (conversion crediting); it had no product ingestion. This adds the catalog half so joined advertiser-campaign products land in `products` and render on `/discover`, mirroring the existing Awin product-feed sync.

## Why

`/discover` currently shows only Shopify (26) + demo-seed (10) products — zero affiliate inventory. Admitad is the closest-to-live network (publisher account verified, Alibaba WW joined, token verified HTTP 200 against `/token/`), and its Products API returns products across **all** connected campaigns via OAuth — no per-advertiser feed-ID discovery needed (unlike Awin).

## Changes

- **`admitad-sync.ts`** — OAuth `client_credentials` (POST `/token/`, HTTP Basic) + paginated Products API fetch + **defensive normalization**: every field tries several candidate names, the raw item is stored on the row, and the first item per campaign is logged. The Admitad product schema isn't pinned from public docs (API is account-gated), so the mapping is confirmed **live on the first activation sync** per the deploy-verification protocol — get a field wrong and it shows in the sample log / `raw` column rather than silently shipping null rows.
- **`providers/admitad.ts`** + registry registration — makes `admitad` a first-class source. `POST /api/v1/internal/sync/admitad` auto-wires via the dynamic provider registry (no route edits).
- **migration** `20260625130000_admitad_marketplace_source.sql` — adds `admitad` to the `marketplace_sources_config.source_network` CHECK constraint (idempotent).
- **`.env.example`** — documents `VCAOP_ADMITAD_*` credential fallback (reuses the Secret-Manager secrets already bound for the postback flow).
- **5 unit tests** locking the defensive field mapping (canonical + alternate names, availability coercion, drop-on-missing, raw preservation).

`tsc` clean; `jest` for the new suite green (5/5).

## Activation (operator steps — not code)

This ships the **plumbing**. Real products on `/discover` still require, in the live gateway:
1. Admitad credentials bound (`VCAOP_ADMITAD_*`, or in the config row).
2. An active `marketplace_sources_config` row (`source_network='admitad'`) with joined `campaign_ids`.
3. First sync via `POST /api/v1/internal/sync/admitad` → **verify the sample-key log + a few `products` rows** before trusting the feed.

A full activation runway (Awin + Admitad) is being drafted separately.

## Verify-at-activation note

The Admitad product field names are defensively mapped but **unverified against a live response**. First real sync must be smoke-tested (sample log + spot-check `products` rows / `raw`) before enabling the daily cron.

---
_Generated by [Claude Code](https://claude.ai/code/session_01SfkEeZCo1sTJFmc1FSRnnK)_

### DIFF
--- a/services/gateway/.env.example
+++ b/services/gateway/.env.example
@@ -50,6 +50,20 @@ SHOPIFY_SYNC_INTERVAL_MS=3600000
 # (Tools -> Postback URL). Bind it as a gateway Cloud Run env / Secret Manager value.
 ADMITAD_POSTBACK_KEY=
 
+# VCAOP Admitad product catalog sync (Products API, OAuth client_credentials).
+# Pulls product data for joined Admitad advertiser campaigns into `products` so
+# items render on /discover. Config lives in the marketplace_sources_config table
+# (source_network='admitad'); these env vars are the credential FALLBACK when the
+# config row omits client_id/client_secret — reuse the same secrets bound for the
+# postback flow. Trigger: POST /api/v1/internal/sync/admitad (or the daily cron).
+# VCAOP_ADMITAD_CLIENT_ID / VCAOP_ADMITAD_CLIENT_SECRET: OAuth app credentials.
+# VCAOP_ADMITAD_BASE64_HEADER: optional pre-built base64("id:secret") Basic header.
+# ADMITAD_API_BASE: defaults to https://api.admitad.com.
+VCAOP_ADMITAD_CLIENT_ID=
+VCAOP_ADMITAD_CLIENT_SECRET=
+VCAOP_ADMITAD_BASE64_HEADER=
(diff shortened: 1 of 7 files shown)


exafyltd:
Closing as superseded: this Admitad catalog-sync scaffold predates a long chain of real Admitad integration work that's already on `main` (title-keyword fallback, default shipping targets, real Bodylab24 deeplinks, keyword-filter fixes — #2856, #2858, #2882, #2883 and more), plus the full Commerce Mesh MVP build (#3061). Rebasing this three-month-old plumbing-only scaffold onto the current marketplace-sync surface would mean re-deriving it from scratch anyway.

---

PR #2808, opened 2026-06-28
fix(orb): Vitana DM follow-ups lose context, mis-route reminders to News

exafyltd:
## 🎯 VTID Reference

**VTID:** `VITANA-REMINDER-NEWS-REDIRECT` _(bootstrap fix; no VTID allocated)_

**Layer:** APIL (gateway conversation bridge)

**Priority:** P1

---

## 📋 Summary

Fixes a reported bug: Vitana proactively surfaced a **Live Room** reminder in the DM chat ("…der Live „Maxina Live-Community Talk" gleich startet. Ich zeige dir die Details."), the user replied "Zeig mir die Details", and the app redirected them to the **News** screen instead of the Live Room.

**Root cause:** `handleVitanaTextReply` in `services/gateway/src/routes/chat.ts` forwarded only the bare user utterance to the conversation/brain layer — with **no thread history and no stable thread id**. Every DM follow-up was therefore processed in isolation. A context-free phrase like "show me the details" gives the navigator no screen signal, so it falls back to a shallow default screen — `/home`, which is the News feed.

The frontend just executes the server-sent navigation directive, so the wrong target is decided entirely in the gateway. Both repos share the branch; the fix is gateway-only (no frontend change required).

---

## ✅ Changes

- [x] Add `fetchVitanaDmHistory()` — loads the recent user↔Vitana DM turns (both directions), oldest-first, excluding the just-inserted current message
- [x] Pass that history as `conversation_history` into **both** the brain (`processBrainTurn`) and legacy (`processConversationTurn`) paths
- [x] Pass a stable per-user thread id (`orb-dm:<userId>`) so follow-ups continue the same thread instead of spawning a fresh, contextless one

With the prior reminder turn in context, Vitana resolves the reference (the named Live Room) and navigates there instead of defaulting to News. This also benefits every other DM follow-up, which previously had no conversational memory.

---

## 🧪 Testing

- [x] Type-checked the changed file (`tsc --noEmit`) — no new type errors introduced (remaining diagnostics are pre-existing missing-dependency/lib errors in this sandbox where node_modules/@types aren't installed)
- [ ] Manual verification in staging: in the Vitana DM, trigger a Live Room reminder, reply "Zeig mir die Details", confirm navigation lands on the Live Room / Live Rooms screen rather than News
- [ ] Confirmed `processWithGemini` builds context only from the passed `conversation_history` (not a separate per-thread store), so the stable thread id does not double-count context

---

## 🔗 Links

- **Related repo:** `exafyltd/vitana-v1` (frontend executes the directive; no change needed)

---

## 🚨 Breaking Changes

None.

---

## 📌 Additional Notes

The navigator (`consultNavigator`) already demotes weak/ambiguous matches to low confidence; this PR fixes the upstream context loss that fed it a signal-less query, rather than re-tuning the scorer. If product wants follow-ups to deep-link to a *specific* room id (not just the Live Rooms list), a follow-up change could persist the room's route in the proactive reminder message metadata.

---
_Generated by [Claude Code](https://claude.ai/code/session_011wkNsTjNYWqM7ETkA7UWbf)_

### DIFF
--- a/services/gateway/src/routes/chat.ts
+++ b/services/gateway/src/routes/chat.ts
@@ -111,6 +111,7 @@ router.post('/send', requireAuth, requireTenant, async (req: Request, res: Respo
       identity.tenant_id!,
       trimmedContent,
       supabase,
+      data.id,
     ).catch(err => console.warn('[Chat] Vitana text reply failed:', err.message));
   } else if (!isVitanaBot(receiver_id)) {
     // BOOTSTRAP-NOTIF-CATEGORIES: Resolve the sender's display name so that the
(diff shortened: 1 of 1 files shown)


exafyltd:
Closing as obsoleted by a later reliability fix: this PR's core change extends the old fire-and-forget `handleVitanaTextReply(...).catch(...)` call in `/send`. That exact pattern was deliberately removed on `main` by VTID-03470 — it was silently losing bot replies under GCP Cloud Run's CPU throttling (the background promise could be frozen mid-write once the request responded). The reply path is now a separate, awaited `POST /vitana-reply` endpoint called explicitly by the frontend. Reintroducing this PR's fire-and-forget extension would regress that fix. If the DM-follow-up-context / reminder-misrouting problem this PR describes is still live, it needs a fresh fix against the current `/vitana-reply` architecture rather than a rebase of this branch.

---

PR #2917, opened 2026-07-22
fix(orb): recover instead of dead-ending on first-connect failure (DEV-COMHU-03406)

exafyltd:
## 🎯 VTID Reference

**VTID:** `BOOTSTRAP-ORB-CONNECT-STUCK` _(no formal VTID exists yet for this fix; tracked under this BOOTSTRAP tag pending one, per repo convention)_

**Layer:** UIUX / voice transport

**Priority:** P0 — production ORB voice reported completely unable to connect

---

## 📋 Summary

User-reported: ORB voice ("Verbinden..." / "Connecting...") stuck forever in production, confirmed not a network issue on their end (strong WiFi, other apps work fine).

Root cause found in `services/gateway/src/frontend/command-hub/orb-widget.js`: the `catch` block that wraps the **first** `session/start` attempt (both the default SSE transport and the opt-in WS transport funnel through the same try/catch) sets an `error` aura on failure (8s client timeout, non-ok backend response, network error) but:
- never calls `_setStatus()` with a new message, so the `"Verbinden..."` text set by `_show()` stays on screen forever — it only gets re-colored **red** via `_setStatus`'s `liveError` fallback, which is exactly the "stuck red Verbinden..." screenshot reported.
- never schedules a retry.

Every *other* disconnect path in this file (SSE `CLOSED`, WS `onclose`) already recovers via `_announceDisconnect()` + `_attemptReconnect()`, which has a proper retry budget, localized "Reconnecting..." status text, and a final "Tap to reconnect" state (`_enterStuckState()`) if retries are exhausted. The very first connection attempt's failure path was never wired into that machinery — a dead end.

Direct production testing (both anonymous and the platform's `e2e-test@vitana.dev` account, against `gateway.vitanaland.com` with a realistic `Origin` header) showed the backend session-start + SSE stream completing in ~1-2s, so this is specifically about **what happens client-side when that first attempt fails** for any reason — the backend healing itself doesn't help a client already stuck on this dead-end path.

---

## ✅ Changes

- [x] `orb-widget.js`: route the initial `session/start` catch block through `_announceDisconnect('connection')` + `_attemptReconnect()` (same recovery path used elsewhere), guarded by the existing `_userInitiatedStop` / `overlayVisible` checks.

---

## 🧪 Testing

- [x] `node --check` on the modified file (plain JS, no build step)
- [x] Manual testing completed — verified prod session/start + SSE stream succeed directly (this fix addresses the failure-handling path, which requires simulating a failed first attempt to fully exercise; not reproduced live since the backend answered successfully in every direct test run during this investigation)
- [ ] Automated tests added/updated — no existing test harness for this widget file
- [ ] Verified in staging/dev environment — pending merge (staging-first cutover: this will auto-deploy to `gateway-staging` on merge to `main`; production requires the PUBLISH button or escape-hatch per repo governance)

---

## 🚨 Breaking Changes

None.

---

## 📌 Additional Notes

Companion cache-bust bump in `exafyltd/vitana-v1` (`index.html` orb-widget.js `?v=` param) so WebView-cached copies of the widget refetch after this ships: see that repo's PR on the same branch name.

---
_Generated by [Claude Code](https://claude.ai/code/session_01P9cP3TApJcMtJGo1edYC7v)_


exafyltd:
Closing as already shipped: the exact bug this PR targets — first-connect failure dead-ending on "Verbinden..." instead of recovering — is already fixed on `main` at the same call site in `orb-widget.js`, via `_attemptReconnect()` wired in with an additional `!_s._isReconnecting` guard this PR's version doesn't have. Different implementation, same fix, already live and slightly more defensive.

---

PR #2922, opened 2026-07-23
Add feature_announcements + admin publish endpoint for News Feed cards

exafyltd:
## 🎯 VTID Reference

**VTID:** none yet — tracked under `BOOTSTRAP-FEATURE-ANNOUNCEMENTS` (same convention as `BOOTSTRAP-PUBLIC-BUSINESS-PROFILE` / `BOOTSTRAP-STAGING-FIRST-CUTOVER`), pending a formal one.

**Layer:** APIL (gateway route) + data (migration)

**Priority:** P2

---

## 📋 Summary

Backend half of the MAXINA News Feed "Brand New Feature" / "Did You Know" cards (frontend: `exafyltd/vitana-v1#922`, `FeatureAnnouncementCard`). Adds a real, admin-published system-post mechanism to replace the frontend's temporary hardcoded demo item, plus the notification fan-out the user explicitly asked for ("let's get everyone notified... in-app and push").

---

## ✅ Changes

- [x] New table `feature_announcements` (migration `20260723150000_bootstrap_feature_announcements.sql`) — one row per announcement, locale-keyed `feature_title`/`description` jsonb, RLS scoped to the caller's own tenant (mirrors `ai_provider_policies`), plus an optional `target_user_ids` column for a **staged test send** to specific users before widening to the whole tenant.
- [x] New admin-only route `POST /api/v1/admin/feature-announcements` (`requireExafyAdmin`, mirrors `admin-notifications.ts`): writes the row, resolves every recipient's locale via `bulkGetUserLocales`, and fans out an in-app + push `feature_announcement` notification per locale group via `notifyUsersAsync`. Pass `recipient_ids` to scope both the row and the notification to specific users only (the test-send path); omit it to reach the whole tenant.
- [x] `GET /api/v1/admin/feature-announcements` — list for admin review.
- [x] `feature_announcement` added to `notification-service.ts` `TYPE_META` (`push_and_inapp`, `p2`, `system`).
- [x] Gateway i18n catalog: `notif.feature_announcement.title` (DE + EN — generic `"Neues Feature: {feature}"` / `"New feature: {feature}"` template; the notification body is the announcement's own per-locale description directly, not a catalog key, since it's one-off editorial copy rather than recurring UI chrome).
- [x] `DATABASE_SCHEMA.md` + `CLAUDE.md` change log updated per governance rule #24.

---

## 🧪 Testing

[...]

PR #2924, opened 2026-07-23
fix(gateway): admin-feature-announcements never verified the JWT

exafyltd:
## Summary

Root cause of why the test-send in #2922's follow-up couldn't authenticate, on both staging and production: `requireExafyAdmin` only *checks* `req.identity` — it must run **after** `requireAuth`, which actually verifies the JWT and populates `req.identity` (its own doc comment says so). `admin-feature-announcements.ts` only had `router.use(requireExafyAdmin)`, so `req.identity` was never set and every call failed `401 UNAUTHENTICATED` — including with a freshly issued, non-expired token carrying `app_metadata.exafy_admin=true`, tested directly against the live production endpoint.

`wallet-admin.ts` has the correct reference pattern: `router.use(path, requireAuth, requireExafyAdmin)`. This PR applies the same fix here.

**Also flagged, not fixed here:** `admin-notifications.ts` (the file this route was modeled on) has the identical gap — worth a follow-up.

## Changes

- `services/gateway/src/routes/admin-feature-announcements.ts`: `router.use(requireExafyAdmin)` → `router.use(requireAuth, requireExafyAdmin)`.

## Test plan

- [x] Confirmed the failure mode directly against production (`gateway.vitanaland.com`) with a real, valid, non-expired admin token before this fix.
- [ ] Will re-verify the same call succeeds once this deploys.

---
_Generated by [Claude Code](https://claude.ai/code/session_01QhBPkg9U1DJyo8ojognubd)_

### DIFF
--- a/services/gateway/src/routes/admin-feature-announcements.ts
+++ b/services/gateway/src/routes/admin-feature-announcements.ts
@@ -29,13 +29,17 @@ import { getSupabase } from '../lib/supabase';
 import { notifyUsersAsync, NotificationPayload } from '../services/notification-service';
 import { bulkGetUserLocales } from '../i18n/server-locale';
 import { tt, type GatewayLocale } from '../i18n/catalog';
-import { requireExafyAdmin, AuthenticatedRequest } from '../middleware/auth-supabase-jwt';
+import { requireAuth, requireExafyAdmin, AuthenticatedRequest } from '../middleware/auth-supabase-jwt';
 import { emitOasisEvent } from '../services/oasis-event-service';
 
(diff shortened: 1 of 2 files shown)


exafyltd:
Fixed in 5b29450 — Codex's suggestion was exactly right: the test mock only exported `requireExafyAdmin`, so `requireAuth` was `undefined` and Express threw constructing the router before any test ran. Mock now provides both, mirroring the real two-stage contract (`requireAuth` sets `req.identity`, `requireExafyAdmin` only checks it).

---

PR #2996, opened 2026-07-30
fix(orb): steer navigate_to_screen away from wrong match-detail screen for "My Matches"

exafyltd:
## 🎯 VTID Reference

**VTID:** `BOOTSTRAP-NOVA-SONIC-VOICE-NAV-FIX` _(follow-up to #2995, same tag — no formal VTID yet)_

**Layer:** APIL (gateway ORB voice pipeline)

**Priority:** P1

---

## 📋 Summary

Follow-up to #2995. After that PR was deployed to `gateway-staging`, live testing surfaced a real repro: asking Vitana "take me to My Matches" made her say "Opening My Matches" — and then nothing happened.

**Root cause (confirmed via `oasis_events`):** the model called `navigate_to_screen` with `screen_id=INTENTS.MATCH_DETAIL` — a single match's *detail* page that requires a real `match_id` — and it failed instantly:

```
Cannot navigate to INTENTS.MATCH_DETAIL: missing required parameter(s) match_id.
```

This never reached the successful-dispatch path #2995 fixed — `dispatchOrbTool` rejected the call before any `orb_directive` existed, so there was nothing to dispatch immediately or otherwise. It's a distinct bug: the tool schema's "HOW TO PICK A screen_id" example list put `INTENTS.MATCH_DETAIL` alongside ordinary list/overview screens with no hint that it's parameterized and needs a real, already-known id — so the model had no reason to prefer the actual matches *list* screen (`COMM.FIND_PARTNER_MATCHES` on mobile / `COMM.MATCHES` on desktop) for a generic "my matches" ask.

**Fix:** added explicit disambiguation guidance to the `navigate_to_screen` tool description — steer generic "my matches" requests to the list screens, and added a blanket "never invent/guess a param value for any parameterized route" rule.

**Separate finding, not fixed here:** the "constantly reconnecting" behavior reported alongside this is a real, pre-existing, and apparently frequent issue — confirmed via `orb.live.diag` events recurring across many sessions over the past week (with or without a preceding tool failure): after some turns the model goes silent long enough (~20s) to trip the response-timeout watchdog, which force-reconnects the session. This is unrelated to #2995 or this fix and needs its own investigation — flagging it here for visibility, not touching it in this PR to keep the change scoped and low-risk.

---

## ✅ Changes

- [x] `services/gateway/src/orb/live/tools/live-tool-catalog.ts` — `navigate_to_screen` tool description: disambiguate "my matches" (list) vs `INTENTS.MATCH_DETAIL` (single, parameterized); never guess a param value
- [x] Updated the two characterization snapshot files that embed this tool description verbatim (`tool-catalog.characterization.test.ts.snap`, `system-instruction.characterization.test.ts.snap`)

---

## 🧪 Testing

- [x] `npx tsc --noEmit` passes clean
- [x] Manually updated both affected snapshots to match the new description text exactly (same escaping convention already used in those files); could not regenerate them via `jest -u` in this sandbox (no `node_modules`, npm registry blocked by egress policy) — **please run the two characterization suites in CI/locally to confirm the snapshots match exactly.**
- [x] Root cause confirmed against real `oasis_events` telemetry (Supabase project `inmkhvwdcuyhnxkgfvsb`) for the exact staging session that reproduced this (`live-02a4f7a0-...`, 2026-07-30)
- [ ] Not yet re-verified live on staging with a real voice ask for "my matches" — recommend doing that once this deploys, and separately investigating the reconnect/watchdog issue noted above.

---

## 🚨 Breaking Changes

None — this only changes prompt/tool-description text sent to the model; no code-path or data changes.

---
_Generated by [Claude Code](https://claude.ai/code/session_01D3z6fKEY5maHGpRfqkQh18)_

### DIFF
--- a/services/gateway/src/orb/live/tools/live-tool-catalog.ts
+++ b/services/gateway/src/orb/live/tools/live-tool-catalog.ts
@@ -2156,6 +2156,9 @@ export function buildLiveApiTools(
             '',
             '── "MY MATCHES" — do not confuse the list with a single detail ──',
             'INTENTS.MATCH_DETAIL is a SINGLE match\'s detail page — it requires a real `match_id` you already have from context (e.g. the user just discussed that specific match). For "show me my matches" / "My Matches" / any general request to see their matches, use COMM.FIND_PARTNER_MATCHES (mobile) or COMM.MATCHES (desktop) instead — those need no parameter. Calling INTENTS.MATCH_DETAIL without a real match_id always fails.',
+            '',
+            '── CONFIRMING A DESTINATION YOU JUST OFFERED ──',
+            'If you just offered to take the user somewhere by name (e.g. you asked "Should I take you to Connected Apps?" or "Soll ich dich zu den verbundenen Apps bringen?") and the user simply confirms without repeating the destination ("yes", "ja", "mach das", "do it", "gerne", "klar", "sure"), call this tool with the SAME screen_id you already resolved for that offer. A bare confirmation carries no screen information on its own — do not treat it as a new, unresolvable request and do not ask again; reuse the screen_id from your own last offer.',
           ].join('\n'),
           parameters: {
             type: 'object',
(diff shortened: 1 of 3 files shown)


exafyltd (commented):
placeholder

---

PR #2997, opened 2026-07-30
fix: Nova barge-in never reached Bedrock; rotation could hard-drop the session

exafyltd:
## 🎯 VTID Reference

**VTID:** `BOOTSTRAP-NOVA-SONIC-VOICE` _(tracking tag used across this Nova voice-provider workstream; no dedicated VTID minted yet)_

**Layer:** APIL (gateway ORB live-session backend)

**Priority:** P0 — both bugs affect every live Nova Sonic voice conversation in production

---

## 📋 Summary

Found during a deep conversation-flow parity audit (Nova vs. Vertex) requested after live reports that Nova "acts stupid" and "the whole conversation logic is gone." Two real, previously-undetected production bugs, both plausibly explaining those reports:

1. **Barge-in is a no-op for Nova.** `NovaSonicLiveClient.sendEndOfTurn()` never sends anything to Bedrock (documented: "Nova uses server-side turn detection... Returning true keeps the session layer's contract satisfied") — it's just a stub. For Vertex, the equivalent call actually tells Gemini to stop generating. For Nova, it doesn't: when a user interrupts, the in-flight response keeps streaming audio, which the gateway then forwards and plays back over the user.
2. **Nova rotation could silently kill the whole session.** Nova's Bedrock stream has an ~8-minute hard cap, so the gateway periodically rotates to a replacement stream (~7.25 min). The rotated stream's rebuilt system instruction is *larger* than a fresh connect's, because Nova has no native session-resumption handle and instead falls back to re-injecting up to ~4000 chars of recent conversation history. Production's `novaClient.connect()` call never set `systemInstructionChunkBytes`, so an oversized rebuilt instruction could trip Bedrock's real `nova_validation` rejection — the exact failure mode the staging bisect harness (`nova-sonic-test-runner.ts`) was built to detect. Rotation deliberately does not retry on failure, so the old stream then rides out to its hard cap and disconnects the user into a brand-new session — matching the "conversation logic is gone" symptom.

---

## ✅ Changes

- [x] `sendEndOfTurn` no-op mitigation: reuse the existing `suppressCurrentTurnAudio` flag (already proven in production for greeting-reemit suppression, auto-clears on the next `turn_complete`) in both the WS interrupt handler (`orb-live.ts`) and the SSE interrupt handler (`live-session-controller.ts`), so audio from a superseded Nova generation is dropped instead of played over the user. No-op for Vertex, which already stops itself.
- [x] New env-tunable `NovaSonicConfig.instructionChunkBytes` (default `4000`, `NOVA_SONIC_INSTRUCTION_CHUNK_BYTES`, `0` disables — same pattern as the existing `keepWarmMs`/`modelWarmMs` fields), wired into the production Nova `connect()` call as `systemInstructionChunkBytes` so oversized rotation instructions get chunked into many bounded `textInput` events instead of one oversized one.
- [x] New config tests for the chunk-size field (env-tunable, explicit-disable, typed invalid-value issue).

---

## 🧪 Testing

- [x] Automated tests added/updated (`nova-sonic-config.test.ts`)
- [x] Full ORB gateway suite green: 140 suites / 2897 tests passed, no regressions
- [x] `tsc` typecheck clean (pre-existing unrelated error in `aws-ecs-admin.ts` — missing `@aws-sdk/client-ecs` — confirmed present on a clean `origin/main` checkout too, not introduced by this change)
- [ ] Verified in staging/dev environment — pending: this needs a live Nova staging session with a mid-response interrupt and a session long enough to trigger rotation to confirm both fixes hold under real Bedrock traffic

---

## 📝 Phase 2B Naming Compliance Checklist

N/A for this PR — backend TypeScript-only change, no new GitHub Actions workflows, no new Cloud Run services/labels, no new files (only edits to existing kebab-case-named files).

---

## 🚨 Breaking Changes

None. Both changes are defensive/additive: the audio-suppression flag reuses an existing, already-shipped mechanism, and `instructionChunkBytes` defaults to a conservative non-zero value but can be set to `0` to fully restore prior behavior if needed.

---

## 📌 Additional Notes

This is one of several findings from a broader Nova-vs-Vertex conversation-flow audit (barge-in/VAD, rotation/context-loss, tool result-shaping across the tool catalog, and wake-brief/journey-guide wiring). This PR ships the two highest-severity, most broadly-impactful fixes (affect every session / every 7+ minute session respectively). Follow-up PRs will address the tool-catalog result-shaping bugs found in the same audit (`resolve_recipient`/`send_chat_message` UUID drop, `view_intent_matches`/`respond_to_match` match_id drop).

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01M5oEMYRbMCpXMFqnQF46eC

---
_Generated by [Claude Code](https://claude.ai/code/session_01M5oEMYRbMCpXMFqnQF46eC)_

### DIFF
--- a/services/gateway/src/orb/live/session/live-session-controller.ts
+++ b/services/gateway/src/orb/live/session/live-session-controller.ts
@@ -2417,6 +2417,13 @@ export async function handleLiveStreamSend(
       // Ungate mic audio
       session.isModelSpeaking = false;
 
+      // VTID-VOICE-NOVA-BARGEIN: see orb-live.ts's handleWsInterruptMessage —
+      // Nova's sendEndOfTurn() never reaches Bedrock, so the superseded
+      // generation keeps streaming audio after this interrupt. Suppress it
+      // via the same flag the greeting-reemit fix uses (auto-clears on the
+      // next turn_complete); no-op for Vertex, which already stops itself.
+      (session as any).suppressCurrentTurnAudio = true;
+
       // Tell Gemini to stop generating
       if (session.upstreamWs && session.upstreamWs.readyState === WebSocketPkg.OPEN) {
         deps.sendEndOfTurn(session.upstreamWs);
--- a/services/gateway/src/orb/live/upstream/nova-sonic-config.ts
+++ b/services/gateway/src/orb/live/upstream/nova-sonic-config.ts
@@ -54,6 +54,23 @@ const DEFAULT_MAX_TOKENS = 4096;
 // bytes keeps each chunk far under the ~32KB range the bisect harness
 // specifically probes as a failure zone.
 const DEFAULT_INSTRUCTION_CHUNK_BYTES = 4_000;
+// PER-TURN LATENCY (Nova voice-provider follow-up): Nova's server-side VAD
+// decides "the user is done speaking" — that decision gates EVERY turn's
+// time-to-first-response, not just the first. Unlike Vertex's numeric
+// `silence_duration_ms` (orb-live.ts `session.vadSilenceMs`, DB-tunable via
+// VOICE_VAD_SILENCE_DURATION_MS), Nova only exposes a coarse HIGH/MEDIUM/LOW
+// `turnDetectionConfiguration.endpointingSensitivity` enum — no numeric ms.
+// This was previously hardcoded in nova-sonic-protocol.ts's `buildSessionStart`
+// default with no way to change it short of a code deploy; `connect()` never
+// even read `options.vadSilenceMs` to inform it. AWS's own guidance: HIGH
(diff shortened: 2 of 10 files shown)


exafyltd:
CI red on `VALIDATOR-CHECK.yml`, but it's failing on the base branch too — checked 10+ consecutive pushes to `main` including the exact commit this PR branches from (`53459ba`), all `failure`, zero jobs scheduled (workflow-level failure, not something in this diff). Every other check (unit, Gateway CI, Test Suite, impact-scan, parity scanner, etc.) is green on the latest commit. Not blocking on this PR; will re-check when the base recovers.

---

PR #2999, opened 2026-07-30
fix: await chat push notification dispatch before responding (BOOTSTRAP-CHAT-PUSH-NOTIF)

exafyltd:
## 🎯 VTID Reference

**VTID:** `BOOTSTRAP-CHAT-PUSH-NOTIF` _(no VTID exists for this yet; tracked under this BOOTSTRAP tag pending one, same convention as BOOTSTRAP-DAILY-FEATURE-TIP)_

**Layer:** APIL (Gateway chat routes)

**Priority:** P1

---

## 📋 Summary

Fixes a bug where **chat message push notifications never arrived externally** (no lock-screen/notification-tray bell), even though in-app delivery worked fine.

Root cause: `chat.ts` (`/send`) and `chat-groups.ts` (`/send`, `fanoutGroupNotifications`, `/refanout-welcome`) all dispatched push notifications via the fire-and-forget `notifyUserAsync()` immediately before sending the HTTP response (`res.status(201).json(...)` / `res.json(...)`). Cloud Run only guarantees CPU while a request is in flight — once the response flushed, the container could freeze/recycle before the still-running `notifyUser()` promise (DB pref lookups, `user_notifications` insert, FCM/Appilix send) finished, silently dropping the push. The `chat_messages` insert and in-app realtime delivery don't go through this code path at all, which is why in-app notifications were unaffected and worked every time.

This is the exact same failure mode already diagnosed and fixed for the daily-feature-tip notification (see `scheduled-notifications.ts`, `BOOTSTRAP-DAILY-FEATURE-TIP`), where the same fire-and-forget pattern caused only 33 of 181 users to receive a push. That fix was never applied to the chat send paths — this PR closes that gap.

---

## ✅ Changes

- [x] `chat.ts` `/send` — switched from `notifyUserAsync()` to `await notifyUser()` before the response, wrapped in try/catch so a push failure can't turn a successfully-sent message into a 500.
- [x] `chat-groups.ts` `/send` — awaits `fanoutGroupNotifications(...)` before the response (was `.catch()`-only fire-and-forget).
- [x] `chat-groups.ts` `fanoutGroupNotifications()` — internal per-member loop switched from fire-and-forget `notifyUserAsync()` to `Promise.allSettled` over awaited `notifyUser()` calls, so the caller's `await` actually waits for every member's push.
- [x] `chat-groups.ts` `/:id/refanout-welcome` (admin endpoint) — same fire-and-forget pattern found and fixed for consistency, since it's already inside a sequential per-member loop.

---

## 🧪 Testing

- [ ] Manual testing completed
- [x] Verified no new TypeScript errors introduced in the edited regions (full `tsc` build blocked in this sandbox by an unrelated pre-existing `tsconfig.json` `moduleResolution` deprecation error, confirmed present on `main` before this change too; `npm install` also blocked by sandbox registry policy — a full build/test pass should be run in CI or by a reviewer with network access)
- [ ] Verified in staging/dev environment

---

## 🚨 Breaking Changes

None — API responses are unchanged; the only behavioral change is that the HTTP response for a chat send now waits for the push notification dispatch (a few hundred ms of DB/FCM calls) to complete before returning, matching the pattern already used for `daily-feature-tip`.

---

## 📌 Additional Notes

Investigated and fixed per user request. No VTID was assigned before starting this fix; recommend allocating one and updating this PR's VTID reference before merge, per governance rules in `CLAUDE.md`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01G8WNmvimt7WKc5e58uQhGb)_

### DIFF
--- a/services/gateway/src/routes/chat-groups.ts
+++ b/services/gateway/src/routes/chat-groups.ts
@@ -25,7 +25,7 @@ import {
   AuthenticatedRequest,
 } from '../middleware/auth-supabase-jwt';
 import { createClient, SupabaseClient } from '@supabase/supabase-js';
-import { notifyUserAsync } from '../services/notification-service';
+import { notifyUser } from '../services/notification-service';
 import { VITANA_BOT_USER_ID, isVitanaBot } from '../lib/vitana-bot';
 import { processConversationTurn } from '../services/conversation-client';
 import { emitOasisEvent } from '../services/oasis-event-service';
@@ -324,14 +324,23 @@ router.post('/:id/send', requireAuth, requireTenant, async (req: Request, res: R
     }
   }
 
-  fanoutGroupNotifications(
-    supabase,
-    groupId,
-    identity.user_id,
-    identity.tenant_id!,
-    fanoutBody,
-    data.id,
-  ).catch(err => console.warn('[ChatGroups] Fanout failed:', err.message));
+  // Awaited (not fire-and-forget) — Cloud Run can freeze/recycle the
+  // container the instant res.status(201) flushes, silently dropping any
+  // still-in-flight background promise (same failure mode confirmed live
+  // and fixed for daily-feature-tip; see scheduled-notifications.ts
+  // BOOTSTRAP-DAILY-FEATURE-TIP and notifyUser's caller in chat.ts /send).
+  try {
+    await fanoutGroupNotifications(
+      supabase,
+      groupId,
+      identity.user_id,
+      identity.tenant_id!,
+      fanoutBody,
+      data.id,
+    );
+  } catch (err: any) {
+    console.warn('[ChatGroups] Fanout failed:', err?.message || err);
+  }
 
   if (trimmed.length > 0 && VITANA_MENTION_RE.test(trimmed) && !isVitanaBot(identity.user_id)) {
     handleVitanaGroupMention(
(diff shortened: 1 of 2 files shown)


exafyltd:
**CI status check-in:** all substantive checks are green (Gateway CI, Services Lint, Test Suite, unit tests, Phase 2B naming/docs gates, ENFORCE-FRONTEND-CANONICAL-SOURCE, Dev Autopilot Impact Scan).

The one failure, `VALIDATOR-CHECK`, is not caused by this PR — it's failing on **every** recent PR in the repo (#2997, #2998, #2962, ...) and even on direct pushes to `main` itself (e.g. run `30530253409`), because it hard-requires a full VTID evidence pack (`docs/validation/<VTID>/{acceptance.md,commands.log,outputs/}` plus `SCOPE_ALLOWLIST:`/`ACCEPTANCE:`/`MERGE_PAYLOAD_PREVIEW:`/`OASIS_IMPACT:` markers in the PR body) that predates this workflow's introduction (2026-07-24) and wasn't required for the precedent PR for this exact class of fix (#2986, merged 2026-07-29, no `VALIDATOR-CHECK` run at all). Since it's systemic and pre-existing, I'm not going to chase it here — flagging for whoever owns that gate to fix or scope down.

---

PR #25, opened 2025-10-29
[VTID DEV-CICDL-0034] Gateway Telemetry — Diagnose, Fix, Validate

exafyltd:
### 📋 Objective
Fix Gateway telemetry endpoints returning 404 to enable event emission from pipelines, agents, and services to OASIS → SSE → Live Console.

---

## ✅ ROOT CAUSE IDENTIFIED

**Finding:** Telemetry endpoints `/api/v1/telemetry/event` and `/api/v1/telemetry/batch` were **never implemented** in the Gateway codebase.

**Evidence:**
- No telemetry router registered in `src/index.ts`
- No `src/routes/telemetry.ts` file existed
- All requests fell through to 404 handler

**Full Diagnosis:** See `docs/reports/devcicdl0034-telemetry.md`

---

### 1. New Telemetry Router ✅
**File:** `services/gateway/src/routes/telemetry.ts`

**Endpoints:**
- `POST /api/v1/telemetry/event` - Single telemetry event
- `POST /api/v1/telemetry/batch` - Batch telemetry events
- `GET /api/v1/health` - Gateway health check
- `GET /api/v1/telemetry/health` - Telemetry subsystem health

**Features:**
- ✅ Validates TickerEvent schema (Zod)
- ✅ Persists to `oasis_events` table
- ✅ Broadcasts to SSE feed (Live Console)
- ✅ Returns 202 Accepted
- ✅ Comprehensive error handling

### 2. DevHub SSE Integration ✅
**File:** `services/gateway/src/routes/devhub.ts`

**Changes:**
- Exported `broadcastEvent()` function
- Telemetry router calls this to push events to SSE
- No breaking changes to existing SSE functionality

### 3. Router Registration ✅
**File:** `services/gateway/src/index.ts`

**Changes:**
- Imported telemetry router
- Registered with Express app
- Updated startup logs

### 4. Tests ✅
**File:** `services/gateway/test/telemetry.test.ts`

**Coverage:**
- Valid telemetry event acceptance (202)
- Invalid payload rejection (400)
- Health endpoint checks (200)

---

## 📊 TELEMETRY EVENT SCHEMA

```typescript
{
  ts: string;           // ISO timestamp (optional, defaults to now)
  vtid: string;         // e.g., "DEV-CICDL-0034"
  layer: string;        // e.g., "CICDL"
  module: string;       // e.g., "GATEWAY"
  source: string;       // e.g., "github.actions"
  kind: string;         // e.g., "deploy.complete"
[...]

PR #3030, opened 2026-08-01
feat(orb): make the WebSocket the browser's default voice transport, and stop the WS session start being a second implementation (VTID-03471)

exafyltd:
## 🎯 VTID Reference

**VTID:** `VTID-03471` _(self-allocated per CLAUDE.md §4.1; `spec_status=approved`, `status=in_progress`)_

**Layer:** APIL (gateway ORB session + WS transport) + the browser widget the gateway serves

**Priority:** P1 — the largest remaining ORB latency item (L-04/L-05), and it closes a silent 7-month feature drift between two session-start implementations.

---

## 📋 Summary

Makes the single bidirectional WebSocket the **default** transport for browser voice sessions, replacing SSE-down + one authenticated `POST` per 64ms audio chunk (~15.6 requests/second while the user is speaking).

**The flip was not the work.** The WS `start` frame ran its own fork of session start, written for VTID-01222 and never re-synced. Everything added to the HTTP/SSE path since had quietly not existed on WS:

| Feature | VTID | Present on SSE | Present on WS (before this PR) |
|---|---|---|---|
| Wake-brief selection | VTID-03079 / 03101 | ✅ | ❌ |
| Journey guidance | VTID-03300 | ✅ | ❌ |
| Guided-topic narration | VTID-03290 | ✅ | ❌ |
| Fast-start wake deferral | ORB-FAST-START | ✅ | ❌ |
| Voice quota gate | VTID-03107 | ✅ | ❌ |
| `AUTH_TOKEN_INVALID` re-auth signal | VTID-AUTH-BACKEND-REJECT | ✅ | ❌ |
| Reconnect continuity (`transcript_history` / `reconnect_stage` / `conversation_id`) | VTID-02020 | ✅ | ❌ |

Flipping the default on top of that fork would have regressed the opener for **every logged-in session** — the exact failure mode that is hardest to notice, because the session still starts and still talks. So the fork is deleted rather than hand-patched a seventh time.

---

## ✅ Changes

- [x] **New `services/gateway/src/orb/live/session/ws-start-adapter.ts`** — runs the WS `start` frame through the same `handleLiveSessionStart` the HTTP route calls. That function only ever touches `req.identity`, `req.headers`, `req.get()`, `req.body` and `res.status().json()`, which is a small enough surface to satisfy from a WebSocket upgrade.
- [x] **~290 lines of forked bootstrap deleted** from `handleWsStartMessage`. What remains is genuinely WS-specific: binding the socket, opening the upstream Live connection, and the `audio_ready` greeting handshake.
- [x] **`cleanupWsSession` now deletes by the LIVE session id.** The live id and the `ws-<uuid>` socket id are different strings now; the old `liveSessions.delete(sessionId)` would have leaked every WS session in the map (and scoped `destroySessionBuffer`/`clearExtractionState` to a key nothing else uses).
- [x] **Session telemetry carries `transport: 'ws' | 'sse'`** instead of a hardcoded `'sse'`, so the two can be compared in one event stream. The wake-timeline recorder gets its own vocabulary (`'websocket'`).
- [x] **Widget default flipped to `ws`**, with three safeguards:
  - automatic **one-shot fallback to SSE** when a WS start fails for a *transport* reason (blocked upgrade, captive portal, proxy that eats 101s), latched per tab in `sessionStorage`;
  - **server rejections are tagged and NOT retried on SSE** — a 401 would be a 401 there too, so retrying just burns another round trip;
  - a **kill switch**: new unauthenticated `GET /api/v1/orb/live/transport`, gated by `FEATURE_ORB_WS_TRANSPORT_ENV`, queried at init. The widget is a static asset served by this gateway, so without it, backing the default out means a redeploy.
- [x] `?v=` cache-buster bumped in the Command Hub `index.html` (companion bump in `exafyltd/vitana-v1`).

---

## 🧪 Testing

- [x] Automated tests added — **26 new** (10 behavioral on the adapter, 16 source-level walls)
- [x] `tsc --noEmit` clean (the two `aws-ecs-admin` errors are pre-existing — `@aws-sdk/client-ecs` is not installed in this sandbox — in a file this PR does not touch)
- [ ] Verified in staging/dev environment — **not done, and it is the gate on this PR. See "Additional Notes".**

**Suite:** full local gateway run is 596/609 suites, 11588 tests passing. The 13 failing suites all fail with `Cannot find module '@aws-sdk/client-ecs'` — the same pre-existing sandbox dependency gap, unrelated to this diff.

**Probed, not just asserted.** Both reversible invariants were checked by reverting the fix and confirming the intended test failed:
- reverted the widget default to `'sse'` → `compiles in 'ws' as the default transport` failed;
- reverted `liveSessions.delete(liveSessionKey)` to `delete(sessionId)` → `deletes from liveSessions by the live session id` failed.

The adapter tests mock `handleLiveSessionStart`, so they assert the request surface directly (identity pass-through, body field forwarding including the five fields the fork dropped, `Bearer` synthesis, `req.get('Origin')` case-insensitivity, no header leakage, no caller mutation).

One updated test: `ws-tenant-fallback.test.ts` pinned the tenant fallback via a string (`isAnonymous: !identity?.user_id,`) that lived in the deleted block. The invariant it protects is still real, so it now pins the new mechanism — the adapter receives the fallback-resolved binding, **and** `resolveOrbIdentity` re-applies the identical fallback inside the controller. Two walls, same invariant.

---

### GitHub Actions
- [x] No workflow files added or renamed

### File Names
- [x] New files kebab-case (`ws-start-adapter.ts`, `ws-start-adapter.test.ts`, `ws-transport-default.test.ts`)
- [x] Directory names unchanged

### Code Standards
- [x] VTID referenced in commit message and in every new code comment block
- [x] Transport values lowercase (`ws`, `sse`, `websocket`)
- [x] No Cloud Run label changes

### Documentation
- [x] VTID in commit messages
- [x] CLAUDE.md changelog row added
- [x] No schema changes, so no `DATABASE_SCHEMA.md` update

---

## 🚨 Breaking Changes

None to the wire protocol — the WS frames are unchanged in both directions, and `session_started` only gains fields (`conversation_id`, and the controller's `meta`: `context_status`, `wake_brief`, `voice_quota`).

Two behavior changes worth naming explicitly:

1. **A WS session start can now be rejected.** A socket presenting an expired JWT gets `{type:'error', code:'AUTH_TOKEN_INVALID'}` and a closed socket instead of silently continuing as an anonymous session with a stranger's greeting. That is the intended fix, but it is a state the WS client could not previously reach.
2. **The live session id in `session_started` is no longer the `ws-<uuid>` socket id.** Nothing in the widget keys off the two being equal, and both server maps are keyed correctly, but anything reading the id out of logs should know the two identifiers diverged here.

---

## 📌 Additional Notes

**This PR should not be merged on my say-so — it wants one real staging session first.** Every claim above is from static analysis and unit tests; I could not open an actual voice session end-to-end from this sandbox (no `GATEWAY_SERVICE_TOKEN`, the same blocker that has #25 / #16 / #12 stalled). The specific things a human should confirm on `preview-gateway.vitanaland.com` with `FEATURE_ORB_WS_TRANSPORT_ENV=staging-only`:

1. an authenticated session starts on WS and the **wake-brief opener is spoken** (the thing the fork never had);
2. barge-in still cuts the model off (the WS path skips the HTTP ordered-send gate by design — a socket is ordered);
3. a reconnect resumes the same `conversation_id`;
4. `context_status: 'pending'` shows up in `session_started` for a cold authenticated start (proves fast-start reached WS).

**Suggested rollout:** ship with `FEATURE_ORB_WS_TRANSPORT_ENV` unset (= `off`), which pins every browser to SSE regardless of the compiled default. Flip to `staging-only`, verify the four points above, then `staging+prod`. Rollback is the same env var, no redeploy.

**Still open, not in this PR:** #25 / #16 / #12 remain blocked on `GATEWAY_SERVICE_TOKEN` — the first real Nova turn-2 latency number still does not exist, and this PR does not produce one either. It removes ~15.6 req/s of HTTP overhead from the upstream path, which is a plausible contributor, but nobody has measured that yet and I am not going to claim it.

**Process note:** `main` requires RepoWise/Graphify queries before implementation. Neither MCP server is available in this session — flagging rather than silently skipping.

---
_Generated by [Claude Code](https://claude.ai/code/session_01M5oEMYRbMCpXMFqnQF46eC)_


exafyltd:
### Parity-scanner drift: 3 of the 13 `high` items are mine, and they are a scanner-scope artifact — not lost observability

The scanner is report-only and all blocking checks are green, but 13 `high` items on a PR that touches the voice path deserves a straight answer about which ones this diff caused. I measured it rather than guessing: ran `npm run extract:ts` against this branch and against `origin/main` in a clean worktree and diffed the extracted topic sets.

| | topics extracted |
|---|---|
| `origin/main` | 22 |
| this branch | 19 |

The three that disappear are exactly:

- `orb.live.context.bootstrap`
- `orb.live.context.bootstrap.skipped`
- `admin.briefing.injected`

The other four `missing_in_vertex` topics (`orb.live.tool.executed`, `orb.navigator.requested`, `orb.navigator.blocked`, `feedback.ticket.created`), all six `watchdog` items, and all 814 `low` items are **already missing on `main`** — pre-existing, untouched by this PR.

**Why the three move, and why nothing is actually lost.** `tools/extract-ts.ts` hard-codes a single file as "the Vertex implementation":

```ts
const ORB_LIVE = resolve(REPO_ROOT, 'services', 'gateway', 'src', 'routes', 'orb-live.ts');
```

This PR deletes the forked WS bootstrap out of that file and routes the WS `start` through the same controller the SSE path uses. All three topics are still emitted — from `live-session-controller.ts` (`admin.briefing.injected` at :921, the bootstrap pair at :988) — so the scanner's single-file view no longer sees an emission that still happens at runtime.

The direction of the change is worth stating plainly, because it is the opposite of a regression: `orb.live.context.bootstrap[.skipped]` was previously emitted **only** by the WS fork, which means the SSE transport — the production default until this PR — never produced them at all. Emitting from the shared controller means both transports now do. That was a deliberate part of this work, not a side effect; the reasoning is in a comment at the emission site.

**Not fixing the extractor here.** Teaching it to follow the delegation into `orb/live/session/**` is a change to a shared CI tool's semantics and would alter the reported surface for every future PR — that wants its own VTID and its own review, not a drive-by edit on a branch already waiting on staging verification. Flagging it as a real follow-up: as more of the voice path moves out of `orb-live.ts`, this scanner will keep reporting false `missing_in_vertex` drift.

Reproduce:

```bash
cd voice-pipeline-spec && npm run extract:ts
python3 -c "import json;print(sorted(x['topic'] for x in json.load(open('extracted/vertex.json'))['oasis_topics']))"
```

---

PR #3031, opened 2026-08-01
fix(watcher): make the Watcher actually learn — wire the distiller, truthful written count (VTID-03531, VTID-03473)

exafyltd:
## 🎯 VTID Reference

**VTIDs:** `VTID-03531` (distiller wiring, backfill, hook contract) · `VTID-03473` (written count) — **Layer:** AGTL · **Priority:** P2

Follow-ups to #3024 (Watcher Phases 1–5), all found by reading the live data and the live docs rather than trusting the deploy. Frontend companion: `exafyltd/vitana-v1#969`.

---

## 📋 Summary

After a week live the Watcher had recorded **1,610 lifecycle steps** including **354 failures**, against a `watcher_lessons` table holding **exactly 0 rows**. It was a recorder, not a memory. Every reminder it could serve came from the 25 hand-authored rules; nothing was ever learned from what actually happened, which is the entire point of the system.

### 1. The distiller had no call site

Phase 2 shipped `distilBatch()` and `upsertLesson()` fully unit-tested. **Nothing ever invoked either one** — zero callers outside tests. Unit tests of pure functions cannot notice that no caller exists, so everything stayed green while the learned half of the system did nothing.

### 2. `upsertLesson()` never incremented `frequency`

A plain PostgREST upsert cannot express `frequency = frequency + 1`, so every lesson sat at 1 forever while `last_seen_at` refreshed on each recurrence. `loadLessons()` withholds frequency-1 lessons until they age past `SINGLETON_QUARANTINE_DAYS`.

**The more often a pattern recurred, the more reliably its quarantine clock got reset.** A recurring failure — the only kind worth remembering — was the one thing guaranteed never to be injected. Fixing #1 alone would have filled the table with lessons that could never be served.

### 3. `writeSteps` reported a hard zero (VTID-03473)

`count: 'exact'` does not compose with `ignoreDuplicates` (`ON CONFLICT DO NOTHING`). Measured: 536 rows genuinely written, `last_written` stuck at `0`, `last_error` null. That field exists *precisely* so "scans every tick, writes nothing" is visible — the one diagnostic built to catch a dead normalizer was blind to exactly that.

### 4. Three session-hook contract bugs

Raised by Codex on the frontend PR, each **confirmed against the [hooks reference](https://code.claude.com/docs/en/hooks)** before fixing. All three were in the copy already merged via #3024, so this repairs live code:

| | Bug | Consequence |
|---|---|---|
| P1 | Read `CLAUDE_SESSION_ID`; no such env var exists — `session_id` is on stdin JSON | The hook **never once** used a real session id; always fell through to `adhoc-<sha>`, merging unrelated sessions |
| P1 | Reconciliation on `Stop`, which fires per assistant *turn* | With `ignoreDuplicates` the **first** write wins — a dirty tree on turn one would **permanently** brand a cleanly-ended session as failed, and that row is what reminders distil from |
| P2 | `rev-list --count HEAD` with no upstream counts all history | Observed live: **68** "unpushed commits" on a branch with one real commit — every reconciliation a false failure |

---

## ✅ Changes

- **`observerTick()` distils** the rows each tick actually **INSERTED** — not those the overlap rescan skipped, which were already distilled when first seen. Idempotent across the rescan with no second dedupe pass. Runs **after both scans**, not between them.
- **`upsertLesson()` matures lessons.** Read-then-write: increments `frequency` **by the number of steps in the batch** (three failures with one signature really are three occurrences), recomputes confidence, merges `evidence_step_ids` bounded at 20, treats `23505` as success.
- **`distilBackfill()` + `POST /distil-backfill`** (admin-gated, `since` required). The tick path only ever sees *new* rows, so the 354 failures recorded before the wiring existed would never become memory — already present, so the rescan skips them forever. This spends that evidence once, through the **same** `distilBatch` + `upsertLesson` path so it cannot diverge from live behaviour.
- **`writeSteps` returns the inserted rows**, not just a count — distillation needs the persisted `id` to link a lesson to its evidence.
- **`GET /health` reports lesson totals** beside cursor state.
- **Session hook**: `session_id` from stdin (jq, `sed` fallback — the no-jq path is deliberate), `SessionEnd` instead of `Stop`, base-ref resolution for unpushed count with **-1 = UNKNOWN** rather than a fabricated number.
- **`WATCHER_REMINDERS_ENABLED=true` on STAGING only**, deliberately **not** in `AWS-PROD-DEPLOY-GATEWAY.yml` — that block is reserved by its own comment for flags with measured prod evidence. Graduates via PUBLISH.

> Note: commit `d9f2712`'s message describes the hook fixes but the commit also carries the backfill work. Same VTID, one branch — flagging it rather than force-pushing a reworded history under a running CI.

---

## 🧪 Testing

- [x] Manual testing completed — verified against live production data
- [x] Automated tests added/updated — **14 new**

**6 wiring tests** that fail if the distiller is ever unwired again — wiring was the gap, so wiring is what they assert:

| Test | Guards against |
|---|---|
| writes a lesson when a tick records a new failure | the shipped bug |
| does NOT re-distil rows the overlap rescan skipped | fabricated recurrences, once per tick forever |
| bumps frequency instead of re-inserting on recurrence | permanent singleton quarantine |
| counts every occurrence in a batch, not one per batch | undercounted evidence; makes backfill correct |
| files an already-recurring pattern as non-singleton | backfill producing lessons quarantine then withholds |
| does not distil successful steps | prompts that argue for cargo-culting |
| merges one tick's repeats | frequency inflated by page count |
| a distillation failure never fails the tick | the degradation posture |

Plus 2 backfill tests, 2 health-counter tests (one asserting an explicit `0` — *absent* reads as "not measured", `0` reads as "measured, and alarming").

Hook corrections verified by execution: `session_id` parsed on **both** the jq and sed paths; unpushed count reads `3` against `origin/main` where it previously read `68`; no-token exits 0 silently; unreachable gateway exits 0.

Gateway suite **613/613 suites, 12,031 tests passing**; `tsc --noEmit` clean.

The observer test fake was extended to return the full inserted **row** rather than a bare id, matching the real client — a fake that models less than the client is how the original gap stayed invisible.

---

### GitHub Actions
- [x] `STAGE-DEPLOY.yml` touched — existing file, staging env list only · [x] N/A · [x] N/A

### File Names
- [x] No new files; all touched files already kebab-case

### Code Standards
- [x] VTIDs in commit messages and code comments
- [x] N/A

### Cloud Run Deployments
- [x] Staging env var added; prod deliberately untouched · [x] N/A

### Documentation
- [x] `CLAUDE.md` changelog rows added for VTID-03531 and VTID-03473

---

## 🚨 Breaking Changes

None. Reminder injection is advisory and flag-gated: `buildReminders()` is wrapped in try/catch at both call sites and a fault yields an empty string, so the worst case is the prompt those paths already produced. It blocks nothing.

---

## 📸 Screenshots

N/A — no UI change.

---

## 📌 Additional Notes

Three things worth carrying forward:

- **`ignoreDuplicates: true` and `count: 'exact'` do not compose** in Supabase. If you need to know how many rows landed, select them.
- **A green unit-test suite proves a function works, not that it runs.** Both distiller bugs were fully covered by *passing* tests. Coverage of the call graph is a different thing from coverage of the functions in it.
- **Verifying a hook's safety contract is not verifying its correctness contract.** The session hook was checked for "exits 0, no-ops without a token" and shipped having never once identified a session correctly. A hook that records the wrong thing looks identical to one that records the right thing until you read the rows.

Remaining and operator-owned:

| Gap | Unblocks | Who |
|---|---|---|
| `WATCHER_SESSION_TOKEN` unset | session ingest (currently `503`, 0 rows) | operator — needs a real secret, not something to commit |
| PUBLISH after staging verification | reminders in prod | your call |

`dev_autopilot_executions` writing 0 steps is **not** a fault — that table's newest row is 2026-07-13; the autopilot has not run in three weeks.

Observer liveness is confirmed independently: ticking every 60s for a week with `last_error` null, which answers the open `setInterval`-on-Cloud-Run question from #3024.


exafyltd:
Three updates, one of which is a correction to a claim in the PR body.

**Correction — the `setInterval`-on-Cloud-Run question is _not_ answered.** The body said a week of unbroken 60s ticks settled it. Merging `main` brought in **VTID-03508**, which set `gateway-staging` to `--min-instances=0`. So the Cloud Run instance is *not* continuously resident, and the continuity almost certainly comes from the **AWS ECS prod** gateway instead, which is always on. Confirmed just now — both stacks are running an observer against the same Supabase project:

| Host | `env` | service |
|---|---|---|
| `preview-gateway.vitanaland.com` | `staging` | `gateway-staging` (Cloud Run, scales to 0) |
| `gateway.vitanaland.com` | `production` | AWS ECS (`cloud_run_service: null`) |

The observer is demonstrably alive; what is *not* demonstrated is that a scale-to-zero Cloud Run service sustains it. Treat that as still open.

**Two writers is safe here, and worth stating explicitly since it wasn't designed for:**
- Steps — idempotent upsert on `(source, source_ref, step)`.
- Distillation — keyed on rows the caller actually **inserted**, so only the instance that won the insert distils. The other gets a conflict-skip and an empty set. No double-counted frequency. That falls out of the VTID-03531 design rather than being special-cased, but it is load-bearing now that two stacks tick.
- Cursor — each writer advances to `max(own maxSeen, cursor)`, so it cannot regress. Worst case is duplicated scan work.
- One real wart: `last_written` reflects whichever instance wrote last, so a healthy value can be masked by the other's `0`. Cosmetic, and `lessons` in `/health` is now the better signal.

**Impact scan warning addressed** (`df0ca6c`) — and I agree with it. `POST /distil-backfill` now emits `vtid.decision.watcher.backfill`. This is not inconsistent with declining the same rule on `/session-step`: a session step is an *observation* and the observer's scan is a poll ("polling is not progress", §6), whereas a backfill is an operator **decision** that changes the memory every later prompt draws on — and it is unreconstructible afterwards, since an upserted lesson looks identical whether it came from a live tick or a backfill. Emitted on the failure path too, because a half-completed backfill is precisely the state someone will need explained.

Also resolved a merge conflict against `main` in the `CLAUDE.md` changelog (pure append on both sides — all 16 rows kept, date-ordered). The conflict is why CI went quiet on this PR for ~20 minutes: with `mergeable_state: dirty` GitHub cannot build the merge commit, so `pull_request`-triggered workflows stop firing entirely. Full suite re-verified against merged `main`: **628 suites, 12,222 tests passing**, `tsc` clean.

Separately, and **not** something this PR caused: `VALIDATOR-CHECK.yml` fails on every run in this repo, including on `main` and on unrelated branches. It requires `VALIDATION_PROFILE:` / `SCOPE_ALLOWLIST:` / `ACCEPTANCE:` / `MERGE_PAYLOAD_PREVIEW:` / `OASIS_IMPACT:` markers in the PR body and restricts changed files to a single-tree allowlist that no multi-area PR can satisfy. It is not a required check. Flagging rather than fixing — shared CI, and it predates this work.

---

PR #3042, opened 2026-08-05
fix(chat): stop the inbox dropping conversations at random (VTID-03493)

exafyltd:
## 🎯 VTID Reference

**VTID:** `VTID-03493`

**Layer:** APIL (gateway route + Supabase RPC)

**Priority:** P1 — user-visible data loss in the inbox, affecting 198 of 220 users

---

## 📋 Summary

Reported as "the messenger chat history is more or less completely gone." Nothing had been deleted — `chat_messages` holds 39,593 rows going back to 2026-02-27. The inbox was *silently dropping conversations at random*.

`get_recent_conversations()` ends its `DISTINCT ON` pass with `ORDER BY peer_id, created_at DESC` — `DISTINCT ON` forces the `ORDER BY` to lead with the distinct key — and then applied `LIMIT p_limit` to **that**. So "the N most recent conversations" was really "N conversations sorted by a random UUID."

With the gateway passing 50, the member with 199 conversations got an arbitrary 50, of which only **7 of their 20 most-recently-active chats** were returned at all. The other 13 were simply absent from the inbox.

---

## ✅ Changes

- [x] `get_recent_conversations()` — wrap the `DISTINCT ON` result (which correctly picks the newest message per peer) in an outer `ORDER BY created_at DESC LIMIT`, so recency selection happens *after* the collapse. Rows now return newest-conversation-first.
- [x] `GET /conversations` — replace the hardcoded `p_limit: 50` with a `limit` query param defaulting to **250** (max 500). 198 of 220 users have more than 50 conversations, so the old value truncated the inbox for almost everyone.
- [x] RPC-failure fallback — it counts *messages*, not conversations, so a chatty peer could eat the whole budget. Now pulls a multiple of the requested conversation count instead of a flat 200.
- [x] `DATABASE_SCHEMA.md` — document `chat_messages` + the RPC, including why the outer query must not be collapsed back into one statement.
- [x] `CLAUDE.md` changelog entry.

---

## 🧪 Testing

Verified against production data (project `inmkhvwdcuyhnxkgfvsb`), using the account with the most conversations (199):

| Check | Before | After |
|---|---|---|
| Of the user's 20 most-recent conversations, how many the RPC returns at `limit=50` | **7 / 20** | **20 / 20** |
| Conversations returned at `limit=250` | — | **199** (all) |
| Result ordered newest-first | no (UUID order) | yes |

- [x] Verified against live data
- [x] `npx tsc --noEmit` clean; `npm run build` passes
- [ ] Automated tests added — **none added.** No test currently covers `routes/chat.ts`'s HTTP surface, and the real defect was in SQL. Worth a follow-up: a fixture with >`limit` conversations asserting the newest ones survive. Flagging rather than silently skipping.
- [ ] Verified in staging

**Migration note:** `supabase/migrations/20260805120000_vtid_03493_recent_conversations_order.sql` is `CREATE OR REPLACE FUNCTION`. It has **already been applied to the production database** — that is how the before/after numbers above were measured. The migration file is the durable record; re-applying it is a no-op.

---

### File Names
- [x] New migration follows the existing `<timestamp>_vtid_<num>_<description>.sql` convention
- [x] No files violate naming canon

### Code Standards
- [x] VTID referenced in commit message and migration header
- [x] Status values / event kinds unchanged by this PR

### GitHub Actions / Cloud Run
- n/a — no workflow or deployment changes in this PR

### Documentation
- [x] VTID in commit messages
- [x] `DATABASE_SCHEMA.md` updated (required: schema/function change)

---

## 🚨 Breaking Changes

None. The RPC keeps its signature and return shape — only row *selection and ordering* change, both strictly toward correctness. The route's new `limit` param is optional; callers that pass nothing get 250 instead of 50.

One thing to be aware of: the inbox response is now up to 5× larger by default for heavy users. Payload is one row per conversation and the RPC still does a single `DISTINCT ON` scan, so this is more rows over the wire, not more queries.

---

## 🔗 Links

- **Companion PR:** exafyltd/vitana-v1#965 — restores per-thread scrollback (39% of all messages were unreachable in the UI)

---

## 📌 Additional Notes

The lesson worth keeping, and the reason this sat undetected: a `LIMIT` after `DISTINCT ON` is not "top N", it is "N arbitrary rows". The function returned plausible-looking data, was named `get_recent_conversations`, and only misbehaved once a user had more conversations than the limit — so it looked correct in every small-account test.

---
_Generated by [Claude Code](https://claude.ai/code/session_016a6iz2aPBYMejdPUQ1r8Gt)_


exafyltd:
## Review finding accepted and fixed — it was real

@chatgpt-codex-connector was right about group rows. Verified against production before changing anything:

```
group rows (receiver_id NULL, group_id set) : 84
clean DMs (receiver_id set, group_id NULL)  : 39,705
receiver_id NULL AND group_id NULL          : 0
receiver_id set  AND group_id set           : 0
```

A clean partition, and the impact was live: across the **16 users** who have sent a group message, `get_recent_conversations` returned exactly **one `peer_id: null` row each** — an inbox entry identifying no DM peer.

The reviewer's framing of the severity is also right. The predicate is unchanged from the original function, but it mattered *more* after this PR: that row used to be buried by the `peer_id` sort, and ordering by recency lets a user's own recent group message surface at the very top.

**Fixed** in the RPC and mirrored in the route's RPC-failure fallback:

```sql
AND m.receiver_id IS NOT NULL
AND m.group_id IS NULL
```

Re-applied to prod and re-verified — null-peer rows across those 16 users went **1 each → 0**.

Also mirrored in the frontend's own `chat_messages` fallback (companion PR), which had the same shape.

### Also in this push

Merged `origin/main` (Polly `#3043` / Bedrock `#3044` had landed; conflict was two changelog rows added at the same spot in `CLAUDE.md` — kept all three). Re-ran `tsc --noEmit` and `npm run build` clean against the merged tree.

---

exafyltd:
## Reopened — the migration needs to land

This was closed without merging at 12:01 UTC with CI 16/17 green and nothing red. Reopening at the repo owner's direction.

The reason this matters more than the code in it: **`get_recent_conversations` has already been changed in the production database**, but with this PR closed, `supabase/migrations/20260805120000_…` isn't on `main`. A live schema change with no migration backing it is a silent trap — anyone rebuilding from migrations, or a later migration that recreates the function, reinstates the original bug, and the symptom is conversations disappearing from inboxes at random with nothing in the repo explaining why.

That is the same shape as the `orb_session_state` incident already recorded in `CLAUDE.md`: a migration that was authored but never applied, undetected for ~2 months because everything failed soft.

### What also isn't shipped while this sits closed

- `GET /conversations` limit 50 → 250 (heavy users still see 50 of their conversations — correctly ordered now, but truncated; 198 of 220 users have more than 50)
- The DM-row filter mirrored into the RPC-failure fallback
- `DATABASE_SCHEMA.md` + `CLAUDE.md` documentation

The companion frontend PR exafyltd/vitana-v1#965 **merged** and is live-bound, so scrollback ships regardless — this is the inbox half.

### Freshened for the current base

`main` moved to `06a4a62` (Titan image gen #3045) while this was closed. Merged it in — clean auto-merge, no conflicts this time. Re-verified against the new base: `tsc --noEmit` clean, `npm run build` exit 0. Head is now `0f75651`.

Nothing about the change itself was altered. If it was closed deliberately — a concern about the 5× larger default inbox payload, or wanting the migration handled another way — say so and I'll adjust rather than re-push the same thing.

---

PR #3049, opened 2026-08-05
fix(orb): collapse the context-bootstrap stampede that hangs ORB on "Verbinden..." (VTID-03504)

exafyltd:
## 🎯 VTID Reference

**VTID:** `VTID-03504`

**Layer:** APIL (gateway) + CICDL (deploy workflow, alert workflow)

**Priority:** P0 — live user-facing ORB voice failure, and it degraded uninvolved users on the same task

---

## 📋 Summary

A user reported (with two photos of an iPhone) that ORB sat on **"Verbinden..."** and then looped **"Verbindung wird wiederhergestellt..."**, never connecting.

It was **not a code regression** — the gateway had been running the same build (`d6861094`) since 08-04, and every piece of infrastructure checked out during the investigation: `session/start` returned in 1.4s, the SSE stream and the WS upgrade both worked, and the deployed widget was byte-identical to `main`.

It was a **self-amplifying retry storm**. Server-side telemetry for `orb.live.context.bootstrap`:

| Window (UTC, 2026-08-05) | p50 | max | over 8s |
|---|---|---|---|
| 04:00 – 11:00 (all day) | ~2.0s | 3.8s | **0** |
| **14:00 hour** | **17.4s** | **119.7s** | **16 of 27** |
| 14:29 (after it drained) | 1.6s | 2.9s | 0 |

with `vtid.live.session.start` + `.stop` pairs repeating every 10–30s — the widget's reconnect loop, recorded from the server. Two users who had nothing to do with the report were dragged to 97s and 119s on the same gateway task.

### Root cause

`system_controls.vitana_brain_orb_enabled` is **TRUE**, so live sessions build their context through `buildBrainSystemInstructionCached` — **not** the legacy `buildBootstrapContextPack` that the prewarm endpoint and most prior tuning target.

That wrapper is gated on `FEATURE_ORB_BRAIN_CACHE`, which `GET /api/v1/admin/feature-flags` reported on prod as `env_var_present: false, setting: "off", live: false`. With the flag off it is a **transparent passthrough** — and what it was passing through was not only the TTL cache but the module's own documented behaviour:

> *"Concurrent-build de-dupe: the in-flight Promise is cached, so prewarm + the tap (or repeated prewarms) share ONE build instead of stampeding."*

So every tap and every reconnect started its own full ~10-Supabase-round-trip build. The builds starved each other, each got slower, the client's 8s start budget expired sooner, and it fired another reconnect. `warmBrainCache()` — the entire purpose of `POST /live/session/prewarm` — was a no-op for the same reason, so the prewarm only ever warmed the path the brain build never reads.

This is the same shape as `BOOTSTRAP-ORB-FASTSTART-DRIFT`: a feature built precisely to prevent a failure, shipped off, never turned on in prod.

---

## ✅ Changes

- [x] **`vitana-brain-cache.ts` — single-flight de-dupe no longer depends on the caching flag.** A stampede guard that only works when an unrelated caching experiment is switched on is not a guard. An in-flight entry is now joined unconditionally, and joined *regardless of TTL* (TTL describes how stale a **finished** build may be; it has nothing to say about one that has not produced a value yet). Flag-off still drops the entry once it settles, so the documented "flag OFF → each call rebuilds" contract is unchanged.
- [x] **`AWS-PROD-DEPLOY-GATEWAY.yml` — pin `FEATURE_ORB_BRAIN_CACHE_ENV=staging+prod`** beside the existing `FEATURE_ORB_FAST_START_ENV`, same rationale and same measured-evidence bar. Task definitions are carried forward verbatim, so a flag that was never copied across the VTID-03419 cutover stays missing forever. Restores 5-min TTL reuse so a retry lands warm instead of paying a fresh ~4.4s build.
- [x] **`buildBootstrapContextPack` — same single-flight (it had none at all), plus bounds on its two unbounded fetch legs.** Only the profiler leg had ever been capped (3s/stage), so the build's true worst case was "however long Supabase takes". The **caller's wait** is bounded at 6s while the build itself runs on to populate the cache — so the retry that follows lands warm instead of starting a third copy.
- [x] **New `ALERT-ORB-BOOTSTRAP-LATENCY.yml`** (hourly) — this incident was found only because someone photographed a phone. Reads `oasis_events` over PostgREST/HTTPS, deliberately **not** `psql "$SUPABASE_DB_URL"` (runner IPs are not on the project's allow-list — VTID-03485/03492). No migration needed.

---

## 🧪 Testing

- [x] Automated tests added/updated — 5 new brain-cache tests (flag-off concurrency, slot release on settle, rejection does not wedge the slot, TTL-outliving build is joined, different users not collapsed) and 3 new bootstrap tests (8-way concurrent collapse, per-identity isolation, retry after failure).
- [x] **Full gateway suite: 624/624 suites, 12156 passed** (7 skipped, 6 todo — pre-existing).
- [x] `tsc --noEmit` clean.
- [x] **Alert workflow verified by replaying the real incident rows through its evaluator**: `p50=27623ms, max=119694ms, 15 builds over 12s → exit 1`. Re-run against the healthy 11:00 window (`p50=2058ms → exit 0`) and an idle window (`no events → exit 0`) to confirm it is not noisy.
- [x] Production behaviour reproduced/measured directly before changing anything: `session/start` 1.4s, SSE stream healthy, WS `101` + `connected` frame, deployed widget identical to `main`, `/api/v1/admin/feature-flags` confirming `ORB_BRAIN_CACHE live:false`.
- [ ] Verified in staging — pending merge (push to `main` auto-deploys staging only, per §16).

### One trap worth flagging for review

A `Promise.race` timeout wrapper placed over a fetch **also swallows that fetch's rejection message**, turning `error: supabase timeout` into a bare `timeout` and losing the cause. The existing suite caught this. The memory leg now keeps its own `.catch` so the original message survives and only a genuine timeout reports one.

---

## 🚨 Breaking Changes

None. Two behaviour changes worth knowing about:

1. **Under overload**, a bootstrap that exceeds 6s now returns to the caller without the memory preamble instead of blocking. That is a real loss of personalization on that turn — and still strictly better than what it replaces, which was a session that never greeted at all because the client had already given up.
2. **A recent-turns fetch failure** now degrades to "no recent-turns block" rather than failing the whole bootstrap, matching how the profiler leg has always behaved. Still logged.

Both are revertible without a redeploy: `ORB_BOOTSTRAP_CALLER_TIMEOUT_MS`, and `FEATURE_ORB_BRAIN_CACHE_ENV` back to `off`.

---

## 📌 Additional Notes

**This does not reach production on merge.** Per §16, merging deploys **staging** only. Production needs the PUBLISH button or `scripts/deploy/publish-to-prod.sh`, and `FEATURE_ORB_BRAIN_CACHE_ENV` only lands on the task definition when `AWS-PROD-DEPLOY-GATEWAY.yml` runs. After promoting, confirm with `GET /api/v1/admin/feature-flags` that `ORB_BRAIN_CACHE` reports `live: true` — note that `isFeatureLive` maps `staging-only` → `isStaging`, so "the var is set" is not the same as "the feature is on".

The storm had already drained by the time of the investigation (14:29 measured 1.6s), so the reporting user should find ORB working again now — but nothing currently prevents a recurrence until this ships.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KAKaoDzUyLoomrtvWpz3wS)_


exafyltd:
CI note: the only check on this PR, `VALIDATOR-CHECK`, is red — but **not because of this diff**. It fails identically on `main` and on every other open branch: **all 30 most recent runs failed**, each one completing in the same second it started with **0 jobs**, which is a workflow *startup* failure rather than a step failure.

Cause: `.github/workflows/VALIDATOR-CHECK.yml` is not valid YAML. Lines 159–165 — the body of a `python3 - <<PY` heredoc inside the "CSP / inline-JS gate" step — sit at column 0:

```
158|              if python3 - <<PY
159|import re,sys
160|p=re.compile(r'''$p''', re.IGNORECASE)
...
165|PY
```

A `run: |` block scalar ends at the first line indented less than the block, so YAML stops the scalar at 158 and then tries to parse `import re,sys` as a mapping key:

```
could not find expected ':'  in ".github/workflows/VALIDATOR-CHECK.yml", line 160, column 1
```

GitHub therefore can't schedule the job at all. The repo's PR governance validator has been inert for at least the last 30 runs.

The fix is to indent 159–165 into the block — YAML strips the block's common indentation before bash sees it, so the `PY` terminator still lands at column 0 and the heredoc keeps working. Verified locally: the file then parses and the reconstructed `run` script still has `PY` alone at column 0.

Not applying it here. It is unrelated to VTID-03504, needs its own VTID, and re-arming a governance gate that has been dead across the whole repo will start enforcing rules on every open PR — that is a deliberate call for a human, not a drive-by fix on an ORB bugfix branch. Flagged separately.

---

PR #3050, opened 2026-08-05
fix(ci): repair unparseable VALIDATOR-CHECK.yml — PR governance was enforcing nothing (VTID-03505)

exafyltd:
## 🎯 VTID Reference

**VTID:** `VTID-03505`

**Layer:** CICDL

**Priority:** P1 — a governance gate that has been silently off repo-wide

---

## 📋 Summary

Found while checking CI on VTID-03504's PR (#3049). `VALIDATOR-CHECK` was red — but red on `main` and on every open branch too. **All 30 most recent runs failed, each completing in the same second it started, with ZERO jobs.**

That is a workflow **startup** failure, not a step failure, and it surfaces as a bare red X with nothing in the Actions UI explaining why. The repo's PR governance validator has been enforcing **nothing** for at least that long.

### Cause

The body of the `python3 - <<PY` heredoc in the CSP gate sat at column 0:

```
158|              if python3 - <<PY
159|import re,sys
160|p=re.compile(r'''$p''', re.IGNORECASE)
...
165|PY
```

A `run: |` block scalar **ends at the first line indented less than the block**, so YAML terminated the scalar at 158 and then tried to parse `import re,sys` as a mapping key:

```
could not find expected ':'  in ".github/workflows/VALIDATOR-CHECK.yml", line 160, column 1
```

Indenting the heredoc into the block is safe and is the correct fix: YAML strips the block's common indentation before bash sees the script, so the `PY` terminator still arrives at column 0, and the deliberately **unquoted** `<<PY` still expands `$p` and `$f` into the Python source. A comment now records the trap so it cannot silently regress.

---

## ✅ Changes

- [x] Indent lines 159–165 into the `run: |` block so the workflow parses.
- [x] Add a comment explaining why the heredoc must stay indented, and what the failure looks like when it isn't (0 jobs, instant red X, no error surfaced).

---

## 🧪 Testing

- [x] File parses as YAML (it did not before): `jobs: ['validate-pr']`, 15 steps.
- [x] Asserted the `PY` terminator is at column 0 **in the script bash actually receives**, after YAML de-indentation.
- [x] **Executed the reconstructed CSP gate against fixtures** — a gate which, given the above, had never once run:
  - clean `.js` → `rc=0`
  - `<div style="color:red">` → `rc=50`, `REJECTED: CSP pattern hit in ... :: /style\s*=/`

---

## 🚨 Breaking Changes

None to any service. But there is a consequence that is the point of this PR, not a footnote:

**Making the workflow parse makes it _run_, and this repo cannot currently satisfy it.**

| Requirement | Reality |
|---|---|
| `VALIDATION_PROFILE:`, `SCOPE_ALLOWLIST:`, `ACCEPTANCE:`, `MERGE_PAYLOAD_PREVIEW:`, `OASIS_IMPACT:` in the PR body (exits 11–15) | `.github/pull_request_template.md` asks for **none of the five** |
| Evidence pack at `docs/validation/<VTID>/` with `acceptance.md`, `commands.log`, `outputs/` (exits 30–33) | The repo contains **exactly one** such directory, `VTID-03237` |
| Path guard: only `services/gateway/{src,dist,openapi}` or the command-hub frontend, plus `docs/validation/` (exit 22) | Rejects any PR touching `.github/`, `CLAUDE.md`, `supabase/`, `scripts/` — **including this one** |

`VALIDATOR-CHECK` is **not** a required status check (#3049 merged with it red), so the effect is a red X on most PRs rather than a merge block.

**The rules were deliberately not weakened to make it green.** Which gate a repo enforces is a governance decision for a human, not something to quietly relax in order to get a green tick. Restoring the parse error is what made the unenforced state visible in the first place; hiding it again behind laxer rules would undo exactly that.

Two coherent follow-ups, either of which is a separate VTID:
1. **Arm it properly** — extend the PR template with the five markers, and add/widen validation profiles so workflow, docs, and migration changes have a legitimate path.
2. **Keep it advisory** — leave it non-required and treat it as a signal, consciously rather than accidentally.

---

## 📌 Additional Notes

Split from VTID-03504 (#3049, merged) rather than folded into it — separate concern, separate VTID, per §4.1.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KAKaoDzUyLoomrtvWpz3wS)_


exafyltd:
`validate-pr` is red on this PR — and that is the fix working, not the fix failing.

**Before:** the run completed in the same second it started, with **0 jobs**, because GitHub could not parse the workflow.
**Now:** the job actually ran — checked out the repo, loaded the ruleset, read the PR title/body, extracted `VTID-03505` — and then rejected on a real rule:

```
REJECTED: VALIDATION_PROFILE missing in PR body
##[error]Process completed with exit code 11.
```

That is the first time this gate has evaluated anything. Every other check on the PR is green (`unit`, `scan`, `check-phase-2b-docs`, `Enforce Phase 2B Naming Standards`).

**I am not making this one green, and it is not possible to without a policy decision.** Adding the five markers to this PR body would move it from exit 11 to **exit 22**, because the path-ownership guard accepts only two profiles:

- `command_hub_frontend` → `services/gateway/{src,dist}/frontend/command-hub/`
- `gateway_backend` → `services/gateway/{src,dist,openapi}/`

both plus `docs/validation/`. **Neither permits `.github/workflows/` or `CLAUDE.md`** — the only two files this PR touches. There is no profile under which a workflow fix can pass. It would also need an evidence pack at `docs/validation/VTID-03505/`; the repo currently has exactly one such directory (`VTID-03237`).

So the gate cannot be satisfied by the change that repairs the gate. Weakening the rules to get a tick would re-hide precisely what this PR exposed, and which rules the repo enforces is your call, not mine — hence the two options in the description (arm it properly with template + profile changes, or keep it consciously advisory).

`VALIDATOR-CHECK` is not a required status check, so this red X does not block merging — #3049 merged with it red for the same reason.

---

exafyltd:
⚠️ Cannot auto-rebase (conflicts likely). Please rebase manually.

Also noting: `validate-pr` is still failing (same known issue documented in the Aug 5 comment — `VALIDATOR-CHECK` cannot pass for workflow/.github changes until a matching profile is added).

---

exafyltd:
**Codex is right, and this PR's description is now wrong. Correcting the record.**

I verified the claim rather than taking it at face value, and it holds:

```
$ git show origin/main:.github/workflows/VALIDATOR-CHECK.yml
180|              if python3 - <<PY
181|          import re,sys          ← already indented on main
...
187|          PY

$ python3 -c "import yaml; yaml.safe_load(open(...))"
main PARSES OK — jobs: ['validate-pr']
```

The indentation repair reached `main` on its own, via **#3059 — "Revive the dead PR governance gate" (VTID-03525/03526/03527)** — a later, independent piece of work that found and fixed the same defect I diagnosed here on 08-05. Two further repairs have landed on top since: `b015650` (path-ownership guard, broken `awk` → `grep -Ev`, VTID-03549) and `f27b244` (fetch base branch so a merge base always exists, VTID-03570).

When `main` was merged into this branch to resolve the changelog conflict, my identical indentation change was absorbed. What is left is comments only — the diff against `main` is **+20 insertions, 0 deletions**, entirely the explanatory block. So the title and body of this PR claim a repair it no longer performs.

**What that leaves:**

- The **fix is done** — by #3059, not by this PR. Nothing is outstanding on the workflow itself.
- The only remaining content here is the regression-guard comment explaining *why* the heredoc must stay indented and what the failure looks like when it isn't (0 jobs, instant red X, nothing in the Actions UI). That is real but minor value, and it is not what this PR says it is.

**Recommendation: close this PR as superseded by #3059.** If the comment is judged worth keeping, it should be re-opened as a small docs-only change with an honest title, not merged under a description claiming to fix something already fixed. I am not closing it unilaterally — the branch now carries a merge commit that isn't mine.

Marking VTID-03505 superseded in the ledger accordingly. Thanks to Codex for catching this; a stale premise surviving into a merged commit is exactly the sort of thing that makes a changelog untrustworthy later.

---

PR #3056, opened 2026-08-06
OASIS: stop the autonomous plane terminalizing session work (VTID-03516)

exafyltd:
## 🎯 VTID Reference

**VTID:** `VTID-03516`

**Layer:** `GOVERNANCE` (worker-orchestrator / vtid_ledger)

**Priority:** P0 — the system of record was asserting that almost all recent work had failed.

---

## 📋 Summary

25 VTIDs carried `status='rejected'` / `terminal_outcome='failed'` for work that is **merged and running in production** — including VTID-03480, whose fix was watched flipping `ok:false` → `ok:true` on live sessions. Rule 1 of `CLAUDE.md` is *"always treat OASIS as the single source of truth"*; nothing ever asserted the source of truth was telling it, so this ran unnoticed for six days.

**The writer, named outright by `oasis_events`:**

```
session allocates VTID, sets in_progress + spec_status=approved   (CLAUDE.md §4.1)
  → ~20-30s later worker-runner-a1846580 claims it
     …because that tuple IS the worker-runner's eligibility predicate
  → GOVERNANCE_CHECK passes 3/3
  → every worker stage fails in <1s:
       "Failed to initialize Anthropic client — ANTHROPIC_API_KEY may be missing"
  → vtid-terminalize        writes terminal_outcome='failed'
  → autopilot-controller    maps that to status='rejected'
```

`autopilot-controller.ts:1047` even says so: *"use 'rejected' for failed tasks (shows red)"*.

Correlation is total: **24/24** affected rows in the last 80 carry **both** a `worker_runner.claimed` event and the ANTHROPIC error, and the autopilot event timestamp matches the ledger's `updated_at` to the millisecond. (`updated_at` is not auto-maintained here — writers set it explicitly — which is why batch-identical values across unrelated VTIDs were real signal.)

**Reproduced live:** VTID-03516, allocated to investigate this, was itself rejected **66 seconds** after being set `in_progress`.

**Ruled out:** DB triggers (only two, both benign — name normalisation and mirror-sync) and `self-healing-reconciler.ts` (writes `status='failed'`, not `'rejected'`, and only touches rows present in `self_healing_log`).

---

## ✅ Changes

- [x] `isAutonomousExecutionTask()` in `routes/worker-orchestrator.ts` — reverses the claim pool from **opt-out to opt-in**. Admits only `metadata.source='self-healing'` (the discriminator four other call sites already key on) or an explicit `metadata.autonomous_execution=true`.
- [x] Enforced on the **pending feed** *and* on the **claim write path** — a worker can call claim with any VTID it likes (stale queue entry, hand-rolled curl, a second poller), so ownership has to be checked on the authoritative write, not only on the read that suggested it. Refusal emits `vtid.decision.claim_refused_not_autonomous`.
- [x] `ci_ledger_integrity_check()` RPC + daily `ALERT-OASIS-LEDGER-INTEGRITY.yml`.
- [x] `scripts/oasis/correct-false-terminalizations.cjs` — repair script, **dry-run by default**.
- [x] `CLAUDE.md` — §4 Task Eligibility gains criterion 5, plus a new "Two execution planes, one ledger" section.

### Why an allowlist and not a denylist

`metadata.source` is **free text**. Sessions write ad-hoc labels describing the task — `orb-voice`, `aws-sns-gchat-alerts`, `news-feed-newest-first`, `gcp-aws-cutover-execution`. Most session-owned VTIDs carry no `claude` marker at all, so a denylist would have caught three strings and kept sweeping everything else.

**Cost of the reversal: zero.** Not one VTID in 60 days carries an autopilot execution link — every VTID the worker-runner claimed in that window was session work it had no business touching.

### ⚠️ Do NOT "fix" this by populating `ANTHROPIC_API_KEY`

The missing key (deliberately deferred on the AWS task defs, §1b) is the only reason each misfire failed in one second. With a working key the worker-runner would instead have begun **autonomously editing code for a VTID a session was concurrently working** — silent, concurrent, conflicting writes, against this repo's own *"Never run parallel VTID executions"*. The eligibility predicate was the bug; the credential was the smoke alarm.

---

## 🧪 Testing

- [x] Manual testing completed — RPC verified against production: **25 findings over 30d, 100% carrying the ANTHROPIC fingerprint**; expected 0 once the gate is live. Workflow `jq` pipeline exercised against real row shapes; repair-script guards exercised.
- [x] Automated tests added/updated — 22 new tests in `worker-orchestrator-autonomous-ownership.test.ts`, including a case pinning that a session VTID satisfies **every** legacy eligibility criterion (which is the whole bug) and is withheld only by the new gate. Full suite: **628/628 suites, 12205 passed**. `tsc --noEmit` clean.
- [ ] Verified in staging/dev environment — pending merge (auto-deploys to staging per §16).

The alert deliberately does **not** fire on `terminal_outcome='failed'` in general. Real failures exist and must stay quiet, or the check gets muted within a week. It fires only on the false-verdict fingerprint: claimed by a worker-runner, terminalized failed, not autonomous-plane work.

---

### GitHub Actions
- [x] Workflow file is UPPERCASE (`ALERT-OASIS-LEDGER-INTEGRITY.yml`)
- [x] No lowercase workflow names present

### File Names
- [x] New files are kebab-case (`worker-orchestrator-autonomous-ownership.test.ts`, `correct-false-terminalizations.cjs`)
- [x] Directory names kebab-case (`scripts/oasis/`)
- [x] No files violate naming canon

### Code Standards
- [x] VTID references are UPPERCASE
- [x] Event type uses snake_case dotted form (`vtid.decision.claim_refused_not_autonomous`)
- [x] Status values lowercase

### Cloud Run Deployments
- [x] N/A — no deploy config touched

### Documentation
- [x] VTID in commit message
- [x] `CLAUDE.md` §4 + changelog updated
- [x] `DATABASE_SCHEMA.md` unchanged — this migration adds a function, no schema change

---

## 🚨 Breaking Changes

**The autonomous claim pool is now opt-in.** Any future producer that expects worker-runner pickup must set `metadata.autonomous_execution = true`; being `in_progress` + `approved` is no longer sufficient. No current producer is affected (self-healing is preserved explicitly, and there is no other autonomous traffic in 60 days).

---

## 📌 Additional Notes

**Ordering — the 25 rows are deliberately NOT bulk-corrected in this PR.** `correct-false-terminalizations.cjs` refuses `--apply` without `--i-have-deployed-the-gate`, because rows returned to a non-terminal state before the gate is live on the gateway the worker-runner polls (AWS prod, `vitana-gateway-awsdr`) would be re-claimed and re-swept, overwriting the very timestamps that identify the writer.

It also does not flip all 25 to `success`, which would replace one false claim with another — *"the autonomous plane had no business judging this"* is not *"this work succeeded"*. Only the **14 with merged commits on `origin/main`** get `success`. The other 11 have their verdict **voided** via `metadata.ledger_verdict_disputed` and are printed for human adjudication; absence of a merged commit is not proof of failure, since some of that work landed in `exafyltd/vitana-v1` and some was investigation with no commit at all.

**Two follow-ups surfaced, not addressed here:**

1. `worker_registry` shows **two** worker-runners heartbeating 11 seconds apart right now (`447859c2`, `a1846580`). VTID-03508 flagged this and could not attribute the second one; *"Never run parallel VTID executions"* cuts against it. One `aws ecs describe-services` call settles it.
2. The handoff document this work was scoped from — `docs/HANDOFF-OASIS-LEDGER-INTEGRITY.md` — **is not in the repository** and the branch it was reportedly pushed to did not exist on the remote. The investigation was reconstructed from the summary in the session prompt; worth knowing that a prior session's push did not land.

---
_Generated by [Claude Code](https://claude.ai/code/session_017trUaGzR2XVaVeq3Vse1TY)_


exafyltd:
### CI status — one real pre-existing failure, not from this PR

**`VALIDATOR-CHECK.yml` is red here, and it is red on every branch.** 8/8 of its most recent runs failed, across four different branches, each with **zero jobs** — the workflow never starts. Not caused by this PR, and not fixed here.

Cause is a YAML block-scalar break in the workflow itself. Inside a `run: |` step, a `python3 - <<PY` heredoc body begins at column 1:

```
158    if python3 - <<PY
159  import re,sys                          ← column 1, escapes the block scalar
160  p=re.compile(r'''$p''', re.IGNORECASE)
```

The block scalar ends at line 159 and YAML then reads it as a mapping key → `could not find expected ':'`. Reproduced locally against `main`:

```
YAML PARSE ERROR: while scanning a simple key
  in ".github/workflows/VALIDATOR-CHECK.yml", line 159, column 1
  could not find expected ':' ... line 160, column 1
```

Worth flagging beyond the mechanics: `CLAUDE.md`'s **VTID-03505** changelog row states this exact problem — *"VALIDATOR-CHECK.yml is unparseable YAML — PR governance silently…"* — was fixed on 2026-08-05. It is still broken on `main` today, so that fix either never landed or was reverted. **PR governance validation is currently dead**, which is the same shape as the bug this PR fixes: a gate emitting a verdict that nothing reads.

Not repaired in this PR — it is unrelated to the ledger-integrity change and warrants its own VTID per §4.1's one-VTID-per-distinct-change rule. Fixing it also means re-enabling a governance gate that has been off for some time, which should be a deliberate, separately reviewed change rather than a drive-by in a governance PR.

### Everything else

The other failures earlier in this PR's history were a GitHub Actions incident, not code. Every one died in `Prepare all required actions` with `Failed to resolve action download info` before fetching any source; three more were cancelled after 3–4h queued having never been assigned a runner (`runner=""`, zero steps, no logs). The repo's Actions queue sat at 0 runners in progress for roughly five hours. All were re-queued and have passed.

Green so far, including the check that actually exercises this change:

- `Gateway CI [VTID: DEV-CICDL-0034]` — the ~7.5k-test gateway suite carrying this PR's 22 new tests. Its own summary step printed `gateway: success`.
- `Services Lint`, `Migration Drift Check`, `PHASE-2B-NAMING-ENFORCEMENT`, `Phase 2B Documentation Gate`, `Dev Autopilot Impact Scan`

Still queued behind the runner backlog: `Test Suite`, `unit`, `OASIS Persistence CI`, `ENFORCE-FRONTEND-CANONICAL-SOURCE`. Locally this branch is **628/628 suites, 12205 tests passed**, `tsc --noEmit` clean.

---

PR #3069, opened 2026-08-09
fix(watcher): make WATCHER_REMINDERS_ENABLED gate every consumer (VTID-03551)

exafyltd:
## 🎯 VTID Reference

**VTID:** `VTID-03551` · **Layer:** AGTL · **Priority:** P1

Found while checking whether the Watcher work had reached production. It had — and prod was reporting the feature **off while injecting**.

---

## 📋 Summary

`GET /api/v1/watcher/remind` reported `enabled_resolved` and then served the reminders regardless. So the flag gated two of its three consumers:

| Consumer | How it reaches the Watcher | Honoured the flag? |
|---|---|---|
| `dev-autopilot-planning` | in-process `buildReminders()` | ✅ genuinely off |
| `dev-autopilot-execute` | in-process `buildReminders()` | ✅ genuinely off |
| **`worker-runner`** | **separate service → this HTTP route** | ❌ **injecting** |

Measured on production with the var unset:

```
enabled_resolved: false
reminders:        6   (3 rules + 3 learned)
```

**A partially-effective kill switch is worse than none**, because it makes you trust a state that is not real — and the state it misreported was *"is this feature touching production prompts right now"*.

### How it survived review

The route's own docstring said:

> In-process callers (the planner, the executor, **the worker-runner bridge**) use `buildReminders()` directly and never traverse this route.

That is false — `worker-runner` is out-of-process and this route is its *only* path. Believing a comment over the call graph is the same failure mode as VTID-03531's unwired distiller: every component individually correct, the seam wrong.

---

## ✅ Changes

The gate keys on **`record_shown`**, which already encoded the caller's intent — no new parameter needed:

| Request | Meaning | Behaviour |
|---|---|---|
| `record_shown=true` | "I am injecting this into a prompt" | Honours the flag; empty bundle when off |
| omitted | "Show me what the Watcher knows" | Always served |

Keeping inspection open is deliberate: an operator needs to see what *would* be injected in order to decide whether to turn the flag on at all. Closing it would make the flag un-flippable-with-confidence.

Three details that are load-bearing:

1. **Suppression happens before `buildReminders()`.** The tests assert `watcher_rules` / `watcher_lessons` are *never queried* — not merely that the array came back empty — so a later refactor that builds-then-filters fails.
2. **Returns 200, not an error status.** `worker-runner` maps any non-2xx to `null`; an error would make a deliberate "off" indistinguishable from a Watcher outage in its logs.
3. **New `injection_suppressed` field.** Otherwise "off, so you got nothing" and "on, and there was genuinely nothing to say" both read as an empty array, and the flag's effect is invisible in the very response it acted on.

---

## 🧪 Testing

- [x] Manual testing completed — the defect was measured on live prod before the fix
- [x] Automated tests added/updated — **4 new**

There were **no route-level tests for `/remind`** before this — only for the pure `buildReminders()`. That is precisely why a route-layer gap went unnoticed while the function it wraps was well covered.

| New test | Guards |
|---|---|
| serves nothing to an injecting caller when off | the shipped bug |
| …and never queries rules/lessons | build-then-filter refactors |
| still serves an inspecting caller when off | the operator preview |
| returns 200, not an error, when suppressed | "off" vs "outage" in worker-runner logs |

Suite **628/628 suites, 12,228 tests passing**; `tsc --noEmit` clean.

---

### GitHub Actions
- [x] No workflow files touched · [x] N/A · [x] N/A

### File Names
- [x] No new files; all touched files already kebab-case

### Code Standards
- [x] VTID in the commit message and the code comment
- [x] N/A

### Cloud Run Deployments
- [x] No deployment change · [x] N/A

### Documentation
- [x] `CLAUDE.md` changelog row + `.env.example` now documents what "off" actually means

---

## 🚨 Breaking Changes

**Behavioural, and intended:** `worker-runner` stops receiving injected reminders in production, because production has the flag off. That is the point — it restores the state prod already believed it was in.

To turn reminders on in prod, set `WATCHER_REMINDERS_ENABLED=true` on the AWS task def. Staging is already `true` via `STAGE-DEPLOY.yml`.

---

## 📸 Screenshots

N/A — no UI change.

---

## 📌 Additional Notes

Third instance of one shape in this subsystem, and worth naming:

| VTID | Defect |
|---|---|
| 03531 | memory nothing wrote to |
| 03534 | memory nothing could read |
| **03551** | **a switch that didn't switch everything off** |

All three passed the full suite throughout. Each component was correct in isolation; the failure was always at a boundary — an uncalled function, a mismatched key, an unchecked flag in the one consumer that lives in another process. Unit tests are structurally blind to all three.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VgXGuAfVfkStTnzoBhnErh)_


exafyltd:
**On the red `validate-pr` check — pre-existing and not caused by this PR.** 17/20 green; the other two are `validate-pr` (×2) and a superseded `scan` (the re-run passed).

`VALIDATOR-CHECK.yml` fails on **every branch in this repo**, including `main`:

```
claude/watcher-agent-plan-pa5y6v    failure
claude/nova-sonic-voice-provider    failure
claude/vtid-03544-gateway-portal    failure
claude/vitana-commerce-mesh-arch    failure
claude/vitana-orb-display-delay-…   failure
```

Failure here is `REJECTED: VALIDATION_PROFILE missing in PR body` (exit 11). I looked at whether I could just satisfy it, and it is **structurally unsatisfiable for a normal PR**: past the body markers it enforces a Path Ownership Guard restricting every changed file to a single tree — `gateway_backend` permits only `services/gateway/{src,dist,openapi}/` and `docs/validation/`. This PR also touches `services/gateway/test/`, `.env.example` and `CLAUDE.md`, so it would be rejected at step 2 even with all five markers added. Any PR that ships a test or updates a doc hits the same wall, which is why it is red everywhere.

It is not a required check — `mergeable_state` is `unstable`, not `blocked`.

Deliberately **not** fixing it here: it is shared governance CI, it predates this work, and repairing it means deciding what the profiles ought to permit — a governance question, not a Watcher one. Flagging it as its own item rather than smuggling a fix into an unrelated PR.

---

